PR #6, opened 2019-12-19
Signal chain uncertainty estimate

mingyuanlu:
This pull request adds features to estimate the systematic uncertainty of A2 & A3 from signal chain efficiency uncertainties, which consist of two parts. They are summarized in the following figure

A23_antennaGain

Our goal is to replace the solid black line, default gain in AraSim, by the dashed line, which is the lower bound of the measured uncertainty in the signal chain squashed into the antenna gain. The two parts, specifically, are:

1. 600MHz low-pass filter. This is to account for the discrepancy between the measured (colored data points) and the simulated gain (solid black line) above that frequency. We interpret this as a lack of understanding of our antenna properties at high frequencies, and so as a lower-bound estimate we kill the gain there. This is implemented by reading an additional low-pass efficiency file when we read the antenna gain files, and scaling the gain values by the low-pass efficiency curve.

2. Efficiency drop of 10% for Vpols and 32% for Hpols in power. These numbers are computed  according to standard error propagation from these individual sources - directivity, transmission coefficient, signal chain noise figure, ambient noise temperature. They are summarized in the table below.

a23_SCErrorEachSource

This is implemented by scaling the antenna gains with an additional factor of efficiency drop (0.9 for Vpols and  0.68 for Hpols). These efficiency drops can be changed using `Settings::SC_EFFICIENCY_ERROR_V` and `Settings::SC_EFFICIENCY_ERROR_H`.

A new `Settings` parameter `USE_SIGNAL_CHAIN_LOWER_BOUND` was added to toggle the use of these features. The default value is 0, where no such lower bound is used. If `USE_SIGNAL_CHAIN_LOWER_BOUND` is 1, then these changes take effect, and the simulated files should be viewed as estimating a lower-bound sensitivity from signal chain efficiency uncertainty.


clark2668:
Awesome! Two quick things:
1) can we update the log.txt file to reflect this addition?
2) Will there be any problems if a user calls `USE_SIGNAL_CHAIN_LOWER_BOUND` without the appropriate other modes? e.g., `DETECTOR=4` and `DETECTOR_STATION=2`, etc. If so, we should add an error check to the Setting.cc file.

mingyuanlu:
1. `log.txt` updated.
2. Do we have more measurements of the signal chain for other ARA stations? If not, I can imagine future analysists wanting to reuse these values as they estimate the systematics (since there is no more data to use). If yes, then sure we should implement the error check.

clark2668:
1. sweet
2. I don't think we have it in the style that Thomas has provided, so these might get reused, especially for stations which are similar (4,5). So I think yes, implement the error check then?

mingyuanlu:
If these values will be reused with stations other than A2 & A3, wouldn't that mean we _don't_ want to implement check against `DETECTOR_STATION`?

On the other hand, checking against `DETECTOR` probably makes sense.

clark2668:
Agreed, I think checking against `DETECTOR` is a good idea.

I think the question is: as the implementation is coded now, will setting `USE_SIGNAL_CHAIN_LOWER_BOUND=1` cause a segfault, or do nothing/something unexpected if `DETECTOR_STATION!=2` or `!=3`. If will break, or do something unexpected/intuitive (like, not be applied at all, and so the user things it's being applied when it's not), then we should stop the user. If it will do something predictable if `DETECTOR_STATION==5`, then I think we're fine. Does that make any sense?

mingyuanlu:
As it's implemented now, `DETECTOR_STATION!=2 & 3 ` will not break or do anything unexpected due to `USE_SIGNAL_CHAIN_LOWER_BOUND=1`. The scaling of the gains is agnostic of which "deep" station is being simulated, since all deep stations read the same set of gain files. In other words, if `DETECTOR_STATION=4` and `USE_SIGNAL_CHAIN_LOWER_BOUND=1`, the simulated gains of A4 will be scaled down in the same manner as A2 or A3.

I've implemented check against `DETECTOR!=4` & `DETECTOR_STATION=0`. The latter due to the fact that the signal chain measurement was done with deep stations (A1 and up), and the TestBed does not apply.

As a side note, currently if one simply sets `DETECTOR_STATION=0` from the default `setup.txt`, a memory corruption is thrown. This is a separate issue that can be opened.

clark2668:
If it doesn't break anything and does what's expected, I say go for it!

And agreed about opening a separate issue, thanks!

mingyuanlu:
Merged. 

The last mentioned issue is opened #7

Issue #7, opened 2019-12-19
DETECTOR_STATION=0 (TestBed) causes memory corruption


mingyuanlu:
Changing DETECTOR_STATION from 2 to 0 (TB) in the default `setup.txt` file causes memory corruption at run time, without any warning or error message regarding incompatible `Settings` parameters.

Error messages at run time that may or may not be related
`Error, station number not match !`
`No string/antenna matches the channel number`
`AraStationInfo::readChannelMapDbIcrr() : statement not prepared OK`

toej93:
It's not clear if the testbed is implemented for `DETECTOR_STATION`. TB has it's own mode: `DETECTOR=3`.


<img width="936" alt="Screen Shot 2019-12-20 at 10 06 33 AM" src="https://user-images.githubusercontent.com/25533803/71280592-7e155e80-2310-11ea-8f20-48bb368504f5.png">
<img width="920" alt="Screen Shot 2019-12-20 at 10 06 03 AM" src="https://user-images.githubusercontent.com/25533803/71280593-7e155e80-2310-11ea-836a-f6c932756b9c.png">

mingyuanlu:
Thanks for tracking that down. So if `DETECTOR_STATION=0` and `DETECTOR` is not 3, the code should stop gracefully, instead of simply segfaults. This can be done simply by checking against such incompatible combinations in the `Settings.cc` section, where other incompatible parameters are checked.

mingyuanlu:
I implemented the check in `CheckCompatibilities` function. Mem error persists. I think this is due to `CheckCompatibilitites` being called _after_ instances of the Detector etc class objects are initialized. And the mem error happens during the instance initialization. 

I think the compatibility check should logically take place right after the setup file is read. No point in creating all the objects only to later find out you have incompatible parameters. As it's currently implemented, the `CheckCompatibilities` function takes an initialized `detector` object as an argument, but only a subset of the checks in this function utilizes it. I propose that we split this function, and keep the checks that need the `detector` object in its current place, while move the other checks to right after reading the setup file. This should solve our current issue, and is more logical from a structural point of view.

Let me know what you think.

toej93:
That sounds good to me. Sounds more efficient than what's currently being done.

toej93:
MYL fixed it and merged to master

PR #8, opened 2019-12-21
Parameter compatibility issue #7

mingyuanlu:
This pull request addresses the issue that AraSim throws a memory error when using `DETECTOR=4` and `DETECTOR_STATION=0`. See issue #7 . 

To solve this, two changes were made:

1. This is essentially a parameter compatibility issue. `DETECTOR_STATION=0` (TestBed) should only be used with `DETECTOR=3`, and no other `DETECTOR` modes. Such check is implemented in the `Settings::CheckCompatibilities` function.

2. The `Settings::CheckCompatibilities` function is called after several class objects are instantiated, eg `Trigger`, `Event`, `Detector` etc. This is probably because the function was implemented to take a `Detector` object as its argument. However, this call placement is illogical, since if there are incompatible parameters, the user should be noticed right after the setup file is read, and the program should stop immediately. As it's implemented, the memory error is thrown in the Detector object instantiation, which comes _before_ the checks. So here I split the `Settings::CheckCompatibilities` function into two functions - one that does not require a `Detector` object (check in point 1 goes here), and one that does. The former is repositioned to be called immediately after the setup files is read. The latter, due to its argument requirement, stays in the same place.

Combing the two changes, the issue is solved. This also safeguards future issues arising from the same code structure problem.

### DIFF
--- a/AraSim.cc
+++ b/AraSim.cc
@@ -83,7 +83,7 @@ int main(int argc, char **argv) {   // read setup.txt file
     settings1->ReadFile(setupfile);
     cout<<"Read "<<setupfile<<" file!"<<endl;
 
-    int settings_compatibility_error = settings1->CheckCompatibilities();
+    int settings_compatibility_error = settings1->CheckCompatibilitiesSettings();
     if (settings_compatibility_error > 0) {
         cerr<<"There are "<< settings_compatibility_error<<" errors from settings. Check error messages."<<endl;
         return -1;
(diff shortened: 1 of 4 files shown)


clark2668:
Awesome, thanks! What if we changed the titles of the functions to "CheckCompatibilitiesSettings" (the one that is called immediately after setup) and "CheckCompatibilitiesDetector" (the one that is called after detector instantiation), so that the function is not overloaded, which I find tends to cause confusion. What do you think?

clark2668:
Also, are we sure we want to commit directly to master? won't this cause conflicts with some of the latest updates from a23-diffuse-4yr?

mingyuanlu:
Sure. I will rename the two functions.

As for the commit plan, I am following the procedure below, which is followed by UW students:
git-model

`hotfix` is branched out from `master`, since this is a ubiquitous problem in all branches I assume. Then according to the procedure `hotfix` should be merged into both `master` & `dev`. We don't really have a `dev`, so I thought it would make sense to merge to `master` first. I can also merge it to `a23_4yr_diffuse` if we want. 

Or do we prefer simply merging to `a23_4yr_diffuse` alone? It will probably be merged back to `master` at some point so I can see it making sense too.

clark2668:
Great, thanks for making that change!

And yeah, I'm somewhat familiar with that git flow. In my head it works better for code that is more modular than AraSim. You're probably right though. But yes, I think go ahead and follow the chart then and make sure the bug-fix get's put into the "develop" branch (e.g., a23_diffuse_4yr) :)

PR #18, opened 2020-04-06
Comparison studies

toej93:
Added Yue's implementation of the frequency-dependent attenuation length model. The following functions were added:
1. `IceModel.cc::GetBessonIceAttenuLength`
2. `IceModel.cc::temperature`
The report class was also changed so this implementation works when USE_ARA_ICEATTENU=2 is used in the setup file. Yue himself tested the implementation when this work was done.

The implementation was compiled and run successfully by myself, so the merge should be straightforward.


clark2668 (changes requested):
Can we rename the depth-dependent function something other than `GetBessonIceAttenuLength`? Just to be more professional?

Also, it seems like there are several changes in `IceModel.h` and `log.txt` that are entirely related to line returns or line ending characters, etc. it would be good if those could be resolved so the "diff" for this pull request is not unnecessarily "noisy"

toej93:
I'll open a new pull request, as I commited a file I didn't want to commit. Maybe this is not the way to do it, but that's the way I know lol

clark2668:
haha, okay

PR #20, opened 2020-04-09
Added a more complete descrption of the variables NNU_THETA and NNU_PHI

toej93:
I only changed `Settings.h`


clark2668 (changes requested):
Can you add a specification for `NNU_D_PHI` is also in radians (presumably). Otherwise I think this looks good.

toej93 (commented):
Done

PR #21, opened 2020-07-14
add option about Z tolerance

MyoungchulK:


clark2668:
Thanks Myoungchul. Can you elaborate more fully on why this issue was fixed? Specifically that requiredAccuracy is the maximum distance the ray solver will tolerate before it stops iterating, and this means that for vertices more than 1km away, the tolerance is a number larger than 1m? And then can you give the links to your 3 talks on docdb so that we have the documentation? thanks!

clark2668:
can you give the links to the docdb slides in the log.txt too?

MyoungchulK:
No problem. I hope here is okay to put the explanation
Let me know my explanation(understanding) is incorrect. I will edit it:)

**Way to get ray solution**

In Ray tracing algorithm, `Earth_to_Flat_same_angle()` in `RaySolver.cc` converts the coordinates of the source (vertex) and target (antenna) into 2 dimentional x and z (x & y coordinates are merged into x coordinate -> (x**2+y**2)^(1/2)). And using Runge-Kupta differntional equation(`doTrace()` in `RayTrace_detail/DoTraceImplementation.h`) for find the solution analitically from source to target. 

In order to perform the `doTrace()`,
First, `traceMax()` and `traceMin()` in `RayTrace.cc` are minimizing the range of ray launch angles that we want to use to search the ray solution.
Second, after setting the range, `traceRoot()` in `RayTrace.cc` is sampling the solutions in the range by using quadratic interpolation and performed the `doTrace()` on each solution until satisfying the condition called `requriedAccuracy`. 

When `doTrace()` is calculating each ray step, the algorithm is forcing to stop the ray when x coordinate of ray reaches in the same x coordinate of the target. And it calculates the differences of z position between ray and target and compares with `requriedAccuracy`. And the `traceRoot()` will conclude the first `doTrace()` that has the differences of z are smaller than `requriedAccuracy` as the ray solution between source and target.

 **requriedAccuracy**

The `requriedAccuracy ` is a parameter that tolerating a certain amount of the z offset between the target and the final ray step. The default is 0.2 m. It corresponds to antenna bandwidth volume. But when the distance between target and source is longer than 200 m, the parameter starts to influenced by the distance. And the `requriedAccuracy` value will become a **Distance/1000 m**. As an example, If the distance between source and target is 5 km, `requriedAccuracy` will become 5 m. It means event though `doTrace()` is resulting in z offset is 5 m, the ray tracing algorithm will allow this solution as a 'right' ray solution. In reality, the 'right' ray solution should be when ray arrives in the antenna volume. It can cause an artifact effect in arrival time delay and vertex reconstruction that baed on timing information between 16 antenna formation. The current `requriedAccuracy` setting is not reflecting reality. The parameter is decided to always limited by **0.2 m**.

**`Z_THIS_TOLERANCE` and `Z_TOLERANCE` option**

In order to fix the issue, the option called `Z_THIS_TOLERANCE` and `Z_TOLERANCE` is added in the `Setting` class. 
So, users can explore the effect of the z offset by themselves.

`Z_THIS_TOLERANCE=0` (default): you can use the original setting that `requriedAccuracy` is changed by **Distance/1000 m**.
`Z_THIS_TOLERANCE=1`: new `requriedAccuracy` setting that you can control the value by setting `Z_TOLERANCE`.

`Z_TOLERANCE=0.2` (default): user can set the `requriedAccuracy` value by whatever they want.

**Effect of the limiting `requriedAccuracy`**

Major two worries about limiting `requriedAccuracy` were effect in simulation run time and losing triggered event.
[Arrival Time Delay in AraSim Part3](http://ara.physics.wisc.edu/docs/0021/002161/001/20200713_Arrival_Time_Delay_in_AraSim_part3.pdf) has detail contents about the worries.
Based on testing the changes, so far, the effect is negligible.

**Related documents in DocBD**

Details are in the 3 talks in below link:

[Arrival Time Delay in AraSim Part1](http://ara.physics.wisc.edu/docs/0021/002152/001/20200622_Arrival_Time_Delay_in_AraSim.pdf)
[Arrival Time Delay in AraSim Part2](http://ara.physics.wisc.edu/docs/0021/002158/001/20200706_Arrival_Time_Delay_in_AraSim_part2.pdf)
[Arrival Time Delay in AraSim Part3](http://ara.physics.wisc.edu/docs/0021/002161/001/20200713_Arrival_Time_Delay_in_AraSim_part3.pdf)

clark2668:
This looks great, and it sounds like you probably have some of the best knowledge of how the ray tracer actually works (re: `doTrace` and `traceMin`, etc.). Thanks for documenting this so thoroughly. I would say go ahead and merge the pull request.

PR #22, opened 2020-10-16
Add files via upload

paramitadasgupta:
The phases are now fixed and the veff as a function of energy is plotted for both new (after the phases were fixed) and old (before the phases were fixed) file. We also show the error in our calculated veff for both the new and old MC files.


toej93:
Hi Paramita,

I don't think we want to add the plots to AraSim. I'd suggest sharing these plots in the discussion of the original pull request.

PR #26, opened 2020-11-04
update antenna loading and add chiba antenna models

clark2668:
This is a rather substantial overhaul of the way antennas are loaded in AraSim.

First, I undid the change made on Jan 23 2020 which introduced the ReadVgainSettings, ReadVgainTopSettings, and ReadHgainSettings. That is PR #13. Instead, I developed a unified set of functions that always uses the settings object.
Also, I eliminated a block functions that was repeated 4 times, and instead bundled them into a single `ReadAllAntennaGains()` function. This make it much more modular.

For better organization, I also moved all of the antenna gain files into a new data/antennas folder.

Finally, I introduced, through ANTENNA_MODE=3, the Chiba XFDTD antenna models.

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -996,25 +996,7 @@ Detector::Detector(Settings *settings1, IceModel *icesurface, string setupfile) 
         
         
 
-	if (settings1->ANTENNA_MODE == 0){
-	  // test read V-pol gain file!!
-	  ReadVgain("./data/antennas/ARA_bicone6in_output.txt", settings1);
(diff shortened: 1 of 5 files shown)


toej93 (commented):
Why don't we keep the Chiba model as the default mode? It seems to be more correct than the previous models

toej93 (approved):
Looks good to me

clark2668:
Great, thanks. I'll squash and merge. And we discussed offline, we'll leave the default ANTENNA_MODE at 0.

PR #29, opened 2020-11-05
Elegantly handle neutrinos with energies outside of cross-section validity range

### LINKED ISSUE #28: E^-2 mode(EXPONENT=2) generates the event that out of the sigma range and makes simulation froze
I'm trying to run the AraSim with 'E^-2' mode(`EXPONENT=2`). But somehow simulation is always stopped with this message (`Need a parameterization for this energy region.(new)`). It happens when `pnu` is outside of the `SIGMAPARAM` in `Primaries::GetSigma`. 
The defined SIGMAPARAM ranges are 1e+13 to 1e+21. But sometimes we are observing `pnu` is over the 1e+21. It caused stopping simulations and not properly closing it. Simulation needs to change to not generate `pnu` outside of sigma range or discard when it happens.
Untitled

clark2668:
This pull request addresses #28. This is an issue where if a neutrino energy is drawn above the range where the cross-section parameterization is defined, the simulation just freezes. Now, the Event class will return an event without an interactions, and AraSim will not proceed in the absence of an interaction. Very verbose as well, which should hopefully reduce confusion.

### DIFF
--- a/AraSim.cc
+++ b/AraSim.cc
@@ -398,6 +398,19 @@ int main(int argc, char **argv) {   // read setup.txt file
 
         // event = new Event ( settings1, spectra, primary1, icemodel, detector, signal, sec1 );
         event = new Event ( settings1, spectra, primary1, icemodel, detector, signal, sec1, Events_Thrown );
+        if(event->Nu_Interaction.size()<1){
+            // If for some reason no interactions were placed into the event holder, continue.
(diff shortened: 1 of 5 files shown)


MyoungchulK (approved):
Nice fix:)

clark2668:
Great, thanks! I'm going to merge this back to the master then.

PR #34, opened 2020-11-20
second commit for in-situ antenna model

MyoungchulK:
In `Report.cc`, all` if (settings1->ANTENNA_MODE == 0 || settings1->ANTENNA_MODE == 3){` ara changed to `if (settings1->ANTENNA_MODE != 1){` to accomodate `ANTENNA_MODE = 4` option into effective height calculation.

In `Detector.cc`, Now `In_situ_VPol_Model.txt` also called by `ReadAllAntennaGains` function for using in `NOISE_CHANNEL_MODE=1` option.


clark2668 (approved):
This all looks good, thanks!

MyoungchulK:
Thank you for the confirm:)

PR #35, opened 2020-11-25
Add sinc interp

clark2668:
This pull request adds a `SincInterpolation` routine to the Tools class. This uses the boost Whittaker-Shannon algorithm (https://www.boost.org/doc/libs/1_71_0/libs/math/doc/html/math_toolkit/whittaker_shannon.html) to do the interpolation.

This method of interpolation is useful for preserving relative spectral power (https://en.wikipedia.org/wiki/Whittaker–Shannon_interpolation_formula). Linear interpolation (the default currently used in AraSim) can artifact the spectrum of a signal at high frequencies. Attached is a plot showing this effect on a neutrino (properties are below), where the Linear interpolator (1) does not preserve the peak amplitude well, (2) adds substantial out-of-band power content to the spectrum, and (3) alters the relative spectral content of the signal by attenuating the high frequencies. The sinc interpolator has much better behavior.

time_domain_comparison

This update does require boost >=1.71.00. For some users, this will require an upgrade. ARA cvmfs trunk is already using boost 1.74.00, so hopefully this will not affect too many users in the production context.

Boost is chosen to do the interpolation (instead of GSL, ROOT, etc.) because AraSim already has boost as a dependency, and boost has a nice interface to the sinc/Whittaker-Shannon method.

Properties of the sample neutrino (in AraSim "read-in" mode):
//evid nuflavorint nu_nubar pnu currentint posnu_r posnu_theta posnu_phi nnu_theta nnu_phi elast_y
00000001 1 1 20.000 0 2585.7312 -0.4323 5.7594 2.2159 8.8515 0.6419

### DIFF
--- a/Report.cc
+++ b/Report.cc
@@ -1033,7 +1033,8 @@ void Report::Connect_Interaction_Detector (Event *event, Detector *detector, Ray
 
 
                                                // skip linear interpolation for now
-                                               Tools::SimpleLinearInterpolation_OutZero( stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt], T_forfft, V_forfft, settings1->NFOUR/2, T_forint, volts_forint );
+                                               // changed to sinc interpolation Dec 2020 by BAC
+                                               Tools::SincInterpolation( stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt], T_forfft, V_forfft, settings1->NFOUR/2, T_forint, volts_forint );
 
                                                // check what we save as V[], volts_forint? or volts_forfft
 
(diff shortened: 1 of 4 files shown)


toej93 (commented):
When I try to compile, I get the following error:
`Tools.cc:10:58: fatal error: boost/math/interpolators/whittaker_shannon.hpp: No such file or directory #include <boost/math/interpolators/whittaker_shannon.hpp>
`
Can this be solved so people don't have to specify where this library lives?

toej93:
Actually, I just realized that the boost that the cvmsf on OSC has doesn't have that folder (possibly and outdated version). I'll install a more updated version in my personal space

clark2668:
Sounds good. But yeah, if you have cvmfs >=1.71.00, and if the `$BOOST_ROOT` is in your `$LD_LIBRARY_PATH`, this should ideally just work.

PR #31, opened 2020-11-16
New pol factor calculation

clark2668:
This factorizes the calculation of the polarization factor into a calculatePolFactor function.  This removes a series of duplicated lines in the ApplyAntFactors* functions.

 Also, this reduces the number of ApplyAntFactors_Tdomain functions. I have eliminated the ApplyAntFactors_Tdomain_Transmitter function by instead passing a useInTransmitterMode  flag to the ApplyAntFactors_Tdomain function itself. This reduces the number of functions lying around.

 Most importantly, it updates the way the polarization factor is calculated to use the theta and phi hat vectors.
 See this PR (https://github.com/ara-software/AraSim/issues/25) and this talk from Ben Hokanson-Fasig (http://ara.physics.wisc.edu/cgi-bin/DocDB/ShowDocument?docid=1947).

PR #38, opened 2020-12-07
update the GetAngleAnt function

### LINKED ISSUE #36: antenna_phi calculation needs an update
The calculation of antenna_phi in the GetAngleAnt function is marked as "incorrect" and needs to be updated.

clark2668:
Update the GetAngleAnt function for a correct calculation of `antenna_theta` and `antenna_phi`. This fixes two things. 

First, GetAngleAnt was not calculating `antenna_phi` correctly.

Second, GetAngleAnt was calculating antenna_theta so that it indicated where the signal was  *going*. But `antenna_theta` needs to indicate where the signal *came from* in order to be used correctly in the new pol factor calculation (see above) and in the antenna gain calculation.

This bug would so far have not impacted AraSim sensitivity estimates because the antennas are approximately azimuthally and zenithally symmetric.

### DIFF
--- a/RaySolver.cc
+++ b/RaySolver.cc
@@ -1432,6 +1432,62 @@ void RaySolver::Solve_Ray (Position &source, Position &target, IceModel *antarct
                     }
 
 		}
+
+		// reorder the solutions (BAC Jan 2021)
+		// it is useful at this point to re-order the solutions so that the direct solution 
+		// (if it exists) is reported first
+		// this copying method is rather inefficient in terms of memory allocation
+		// but also, it's not obvious that it ever gets called 
+		// (Brian couldn't find a case in testing on a few hundred events)
+		// so probably it doesn't matter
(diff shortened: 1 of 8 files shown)


MyoungchulK:
👍

clark2668:
Okay, I'd like to get this merged back into AraSim. I ran a comparison of the latest in this branch (d9725c2) again a version of the trunk from a while ago (4108ec3). I chose this comparison specifically because it pre-dates the polarization calculation change (#31, 79a7fe3), which we know is sensitive to a bug in the phi calculation, which this PR fixes. See issue #36. 

As you can see in the plot, things look good at the veff level (differences are essentially within error bars). Are we good to merge? @toej93 ?

compare

toej93:
Yea, go for it

clark2668:
Great, thanks! I'll have to resolve that last conflict and then merge down.

PR #40, opened 2020-12-22
Adding solution for avoiding throw command in Raytracing

### LINKED ISSUE #37: Simulation is crashing by TraceFinder::refineRoot
Sometimes simulation is crashing by this message. 
Trying to figure out what is causing it.
Untitled

MyoungchulK:
In the very far vertex case (~12km), the `refineRoot` and `evenmore_refineRoot` function can't find the optimal launch angle that satisfying `requiredAccuracy`.
In this case, the function executes the `throw` command, and the simulation is aborted.
In order to avoid the aborted, command out the `throw` and returning the closet results (`refine_result`) for 2nd direct ray tracing.
If the closet result (`refine_result.second`) is bigger than the missing_z from the 1st doTrace (`results.front().miss`), It will **return 0 ray solution**.
If the closet result(`refine_result.second`) is smaller than the missing_z from the 1st doTrace or `requiredAccuracy`, 2nd direct ray tracing will try to find the solution with it(`est.angle`).

The detailed story is in [issue#37](https://github.com/ara-software/AraSim/issues) and this [slide](http://ara.physics.wisc.edu/docs/0022/002263/001/20201221_Raytracing_Bug.pdf).

I will complete this pull request by adding the **SVAE_RAY_PATH** option. Probably tomorrow:)


clark2668:
I assume you checked that events still pass and that the ray tracer didn't break? In which case, also should be okay, go ahead and merge.

MyoungchulK:
Yup, they didn't break even though I set the radius of the volume to 12km haha.
I put the **SVAE_RAY_PATH** option into `Setting.cc & h`. But I didn't apply to `Report.cc`
I will merge it after fully implement it into the Report class.

clark2668:
Okay, very good. And okay, once you fully have this ready to go, let us know.

MyoungchulK:
Now, the say steps between the vertex (source) and an antenna (target) are saved in the [ray_step](https://github.com/ara-software/AraSim/blob/58a248ad409a4c30ff23b83b7761870d793306b9/Report.cc#L444) array, unless `DATA_SAVE_MODE` is 2.
SVAE_RAY_PATH option is replaced by DATA_SAVE_MODE

clark2668:
Great, thanks. The diff on the Report.cc file looks _enormous_. It would seem virtually every line has been changed. I suspect your IDE/code editor changed the tabbing format? We need to fix this I think, such a large diff would be terrible for the history.

One other questions: I don't see the SAVE_RAY_PATH change? Or is it hiding?

MyoungchulK:
> Great, thanks. The diff on the Report.cc file looks _enormous_. It would seem virtually every line has been changed. I suspect your IDE/code editor changed the tabbing format? We need to fix this I think, such a large diff would be terrible for the history.
> 
> One other questions: I don't see the SAVE_RAY_PATH change? Or is it hiding?

MyoungchulK:
> > Great, thanks. The diff on the Report.cc file looks _enormous_. It would seem virtually every line has been changed. I suspect your IDE/code editor changed the tabbing format? We need to fix this I think, such a large diff would be terrible for the history.
> > One other questions: I don't see the SAVE_RAY_PATH change? Or is it hiding?

Oops. accidentally closed it let me reopen it

MyoungchulK:
> Great, thanks. The diff on the Report.cc file looks _enormous_. It would seem virtually every line has been changed. I suspect your IDE/code editor changed the tabbing format? We need to fix this I think, such a large diff would be terrible for the history.
> 
> One other questions: I don't see the SAVE_RAY_PATH change? Or is it hiding?

I incorporated `SAVE_RAY_PATH` into `DATA_SAVE_MODE` option. So, AraSim will automatically save ray steps for all events (or triggered events) unless `DATA_SAVE_MODE = 2`, since option 2 is designed to save only physics values.
If we think it will increase the output size, I can just keep the original `SAVE_RAY_PATH` option.

MyoungchulK:
Probably it is hard to see since Report changes are somehow big. Let me work on that first!

MyoungchulK:
I fixed the diff. on the `Report.cc`. So. It is easier to check my implementation.

clark2668:
Okay, I think this looks okay! Please squash and merge.

PR #45, opened 2021-03-03
Add ffttools to replace the AraSim FFT routines

clark2668:
This update to AraSim makes it so that the AraSim FFT routines are now based on FFTW through the FFTools package and libRootFftwWrapper (https://github.com/nichol77/libRootFftwWrapper).

Special care was taken to make sure that the data packing, etc. did not change in `realft`. So this operates a drop-in replacement.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,8 @@ ifeq ($(strip $(BOOST_ROOT)),)
 	BOOST_ROOT = /usr/local/include
 endif
 
-SYSINCLUDES	= -I/usr/include -I$(BOOST_ROOT) -I$(PLATFORM_DIR)/include
-SYSLIBS         = -L/usr/lib -L$(PLATFORM_DIR)/lib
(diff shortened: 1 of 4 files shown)


cozzyd:
Did we ever check that the difference in FFT sign convention doesn't cause any issues?

clark2668:
I thought Yuchieh checked this pretty carefully: https://aradocs.wipac.wisc.edu/cgi-bin/DocDB/ShowDocument?docid=2313

Do we see a problem?

cozzyd:
ah great, I missed that comparison. We had issues on PUEO doing the same thing because there were some transfer functions defined in the frequency domain that then became backwards, but it seems like there are no such problems here based on that presentation...

clark2668:
Yeah, no problem. I think we have good evidence things are working alright, but if not, would definitely be good to know!

PR #48, opened 2021-03-14
Add a GitHub workflow with a veff test

clark2668:
This will add a GitHub workflow to the AraSim repository that will run a effective volume test.

So, on every push to the master, or pull request to the master, the veff test will run. First it will try to compile AraSim. Then it  will run 500 neutrinos at 1 EeV with a "vanilla" ARA station set up file. It will check that the output file is created, that it contains the AraSim internal check for completion. Then it checks that 2 events pass with a specific summed weight. Because the seed of the setup file is fixed, this test should be strictly reproducible. More tests can be added later.


bhokansonfasig (commented):
With my limited AraSim experience I'd say this looks good. Definitely will be a useful tool to avoid major breaking changes in the code! Also looks like it provides a good base for potentially adding more tests in the future.

My only suggestion would be that the `check_veff.py` file could be cleaned up a bit. It's probably cleaner to use `sys.exit` to break on a failing subtest as it happens, rather than propagating the `pass_*_test` booleans through the file. That would make it easier to add or take away subtests too as necessary.

clark2668:
Yeah, that's a good point. What I've done isn't very extensible. Let me change that...

bhokansonfasig (approved):
Looks good!

toej93:
Looks good to me too

toej93 (approved):

PR #54, opened 2022-07-01
Update the noise modeling in AraSim for deep stations

clark2668:
This pull request is a major upgrade to how noise is modeled/handled in AraSim for deep stations.
In particular, this pull request enables the user to load Rayleigh noise parameters and use them to simulate noise in the deep stations.
It also involves some major cleanup of the associated functions in the Detector and Report classes.


clark2668:
Let me also add one comment for context: the diff on the Report.cc file is extremely large on purpose. I linted the Connect_Interaction_Detector function, because it's tabbing was out of control and impossible to read. This generates an unfortunately large diff, but I think we have to bite the bullet on this for readability and future usability.

PR #58, opened 2022-11-02
Fixed PickExact function to be consistent with usual coordinate conventions.

Machtay:
The function PickExact in the Primaries class was using a different convention for the coordinate position of events than the usual one in AraSim (theta was an elevation instead of a zenith). I detailed the changes in a comment above the function and I'll add a description to the wiki.


clark2668:
This mostly looks good! A question and a request:

Question: Does this affect the convention used in event read in mode, supported by the [EVENT_GENERATION_MODE](https://github.com/ara-software/AraSim/blob/master/Settings.h#L288) function? That is, does event read in mode use PickExact? If so, then we need to make sure the docs there are updated.


Request: please remove the `sigmaCrossSection.pdf` file from the commit.

Machtay:
@clark2668 whoops, thought I deleted that. Should be gone now!

Yes, I just checked and it looks like [EVENT_GENERATION_MODE=1](https://github.com/ara-software/AraSim/blob/30012a05aeabe83aeffb85906194ea2729a7a035/Event.cc#L41) does use INTERACTION_MODE=2 (PickExact). I don't think we would need to change anything more in AraSim, but this does mean that old [event files](https://github.com/ara-software/AraSim/blob/30012a05aeabe83aeffb85906194ea2729a7a035/Settings.cc#L692) would need to be modified to change the position convention to match this one.

clark2668:
Thank you! Okay, would you mind fixing up the eventReadIn.txt file (https://github.com/ara-software/AraSim/blob/master/eventReadIn.txt) so that it has the correct coordinates?

clark2668:
Thanks Alex! I suspect you are now clear to merge.

clark2668:
Oh, can you please add a note to log.txt that you also changed the event read in example? THEN you're ready to merge :)

PR #59, opened 2022-11-20
Added a basic model for simulating the Greenland Attenuation Lengths.…

FlahertyJustinC:
…  User must set GREENLAND_ATTEN in their setup.txt, where 0 (default) just uses the current South Pole models, whereas 1 and 2 use the upper and lower limits of the Bogorodsky model in the RNO-G publication linked here: doi.org/10.1017/jog.2022.40.  It simply replaces the South Pole attenuation lengths with the Greenland values.


clark2668:
Thanks Justin! Is there a reason this can't be handled by just changing `ICE_MODEL`? Does it need to be it's own flag? I guess you're copying the `MOOREBAY` parameter?

clark2668:
Also, as this model looks frequency independent, I advise adding a check in Settings.cc to prevent the use of `GREENLAND_ATTEN!=0` and `USE_ARA_ICEATTENU=1`. In other words, make sure that when you're using the Greenland model, AraSim is using the freq-independent parts of the code.

IceModel.cc
```diff
@@ -171,6 +171,31 @@ void IceModel::setUpIceModel(int model) {
...
+  while(greenlandUpper>>d_greenlandUpper[i]>>l_greenlandUpper[i])
+    {
+      i++;
+    }
+  greenlandUpper.close();
+    
+  ifstream greenlandLower("data/greenland_attenlength_bogorodsky_lower.txt");
+  if(greenlandLower.fail())
+    {cerr << "Failed to open greenland_attenlength_bogorodsky_lower.txt";
+      exit(1);
```
clark2668, line 191: Does `exit(1)` correctly kill the program? That is, will it terminate all of AraSim? I prefer `throw std::runtime_error`, but it's a matter of taste. The important thing is that it stops AraSim.

clark2668 (changes requested):
This for the most part looks good. I asked a few questions above. Also, can you add a comment to log.txt?

PR #62, opened 2023-02-12
Per-channel gain implementation

AlanSalcedo:
This is pull request is an upgrade of gain modeling. It allows for per-channel in situ modeling.


Detector.cc
```diff
@@ -2088,14 +2088,30 @@ Detector::Detector(Settings *settings1, IceModel *icesurface, string setupfile)
 	    cout<<"start read elect chain"<<endl;
         if(settings1->CUSTOM_ELECTRONICS==0){
             //read the standard ARA electronics
-            cout<<"     Reading standard ARA electronics response"<<endl;
-             ReadElectChain("./data/ARA_Electronics_TotalGain_TwoFilters.txt", settings1);
-            //ReadElectChain("./data/ARA_Electronics_TotalGainPhase.txt", settings1);
-        }
+            if(settings1->DETECTOR_STATION==2 || settings1->DETECTOR_STATION==3){
```
marcomuzio: Is it possible to make this if statement condition be something like `if(settings1->DETECTOR_STATION != default_value)`? That way this won't need to be changed by hand as more stations' gain models are added.

Detector.cc
```diff
@@ -4116,186 +4132,297 @@ inline void Detector::ReadCalPulserWF(string filename, Settings *settings1 ) {
...
-            N++;
+	// check if the gain file exists
+	char errorMessage[400];
+	struct stat buffer;
 
-            xfreq_tmp.push_back( atof( line.substr(0, line.find_first_of(",")).c_str() ) );
-            // cout<<"freq : "<<xfreq_tmp[N]<<"\t";
+	bool gainFileExists = (stat(filename.c_str(), &buffer)==0);
+	if (!gainFileExists){
+		sprintf(errorMessage, "Gain model file is not found (gain file exists %d) ", gainFileExists);
```
marcomuzio, line 3812: Maybe it's useful to return the `filename` that `ReadElectChain` was passed if this error is thrown. Could also be added to some of the error messages below too possibly.

Detector.cc
AlanSalcedo: A few lines earlier (L2075) Brian wrote `if(settings1->DETECTOR_STATION > 0)` for the noise to exclude the Testbed. I could do this? I was just concerned about being able to simulate A1 now with the standard gain model. This wouldn't be a problem when gain models are added for A1.

Report.cc
```diff
@@ -387,7 +387,18 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
 
                 stations[i].strings[j].antennas[k].clear(); // clear data in antenna which stored in previous event
 
-                // run ray solver, see if solution exist
+		// This (gain_ch_no) is used for per-channel gain implementation. 
+		// It is used in all instances of ApplyElect_Tdomain() and ApplyElect_Tdomain_FirstTwo(), to indicate channel number
+		// Note that channel numbering is different for DETECTOR==4 than for the other modes (1-3). See that in the definition of GetChannelfromStringAntenna()
```
marcomuzio, line 392: Would it be better to handle this inside `GetChannelfromStringAntenna`? That way it always returns what it's purported to? Unless this would break something else. Similarly at L2816.

Settings.cc
```diff
@@ -974,13 +974,22 @@ int Settings::CheckCompatibilitiesSettings() {
 	        num_err++;
 	    }
         if(DETECTOR_STATION_LIVETIME_CONFIG>-1){
-            if((int)DETECTOR_STATION==2 || (int)DETECTOR_STATION==3){
+            if((int)DETECTOR_STATION==2){
                 cerr<<"DETECTOR_STATION_LIVETIME_CONFIG is set to "<<DETECTOR_STATION_LIVETIME_CONFIG<<endl;
```
marcomuzio: Why is this an error message? At this point do we know an error has occurred? Sorry I know this may not be something you added.

Settings.cc
```diff
@@ -974,13 +974,22 @@ int Settings::CheckCompatibilitiesSettings() {
 	        num_err++;
 	    }
         if(DETECTOR_STATION_LIVETIME_CONFIG>-1){
-            if((int)DETECTOR_STATION==2 || (int)DETECTOR_STATION==3){
+            if((int)DETECTOR_STATION==2){
```
marcomuzio, line 986: Is there a more scalable way to hand these kinds of error messages?

Report.cc
AlanSalcedo: Unfortunately, people have used this function in many places:

For instance, in L2171 someone wrote: `int channel_num = detector->GetChannelfromStringAntenna(i, string_i, antenna_i, settings1);`

Then, used channel_num in L2191 and manually shifted it by "-1":

`if (trigger->Full_window[trig_j][trig_i + trig_bin] < (detector->GetThres(i, channel_num - 1, settings1) *trigger->rmsdiode *detector->GetThresOffset(i, channel_num - 1, settings1)))`

If I shift the numbering in `GetChannelfromStringAntenna`, all these instances will be shifted by an extra "-1".

data/gain/README.txt
marcomuzio, line 1: Would you mind adding a reference to the technical note for more information re: formatting etc.

Detector.cc
marcomuzio: Maybe a solution would be to have `ReadElectChain` switch to the default model (TotalGain_TwoFilters.csv I think?) if it find that `the_gain_filename` doesn't exist. And spit out an error message warning that it couldn't find the requested file and is falling back to the default.

Report.cc
marcomuzio: I see, but do those shifts have to do with station 4? I don't see an if statement that only applies it for that station but maybe I'm missing it. If it's applied to all stations, I would think it should be safe to have `GetChannelfromStringAntenna` take care of the special case for station 4 internally.

Settings.cc
AlanSalcedo: You're right. I just tested it and we can get rid of this. The next error takes care of the same.

Report.cc
AlanSalcedo: It's not the station but the simulation mode. Mode 4 is for real deep stations. The numbering there goes from 1-16 and people shifted it to 0-15 to point to columns in arrays. I think that if I fix this in `GetChannelfromStringAntenna`, I would have to track all instances where people made shifts manually, like in the example above, and made sure things work. 

My implementation would work fine if I fix the function. But people did shifts in other places. i.e:

if (trigger->Full_window[trig_j][trig_i + trig_bin] < (detector->GetThres(i, channel_num - 1, settings1) *trigger->rmsdiode *detector->GetThresOffset(i, channel_num - 1, settings1)))

and those would now be shifted twice

Detector.cc
AlanSalcedo: Ok, I'll think what's the best way to do that. We'll also want it to fall back to default noise in that case, right?

marcomuzio: Ya I think that's right. It shouldn't crash things it should just throw an error message to warn the user.

Report.cc
marcomuzio: Sorry maybe I'm missing something. The condition is on the station number not the mode right? Maybe they are the same, I'm not familiar enough with AraSim. @clark2668 maybe you have thoughts? Agreed that if it will break things elsewhere it's better to just leave it as is.

Settings.cc
marcomuzio: Awesome thanks!

Detector.cc
AlanSalcedo: I fixed it according to your first comment and used this instead: `if(settings1->DETECTOR_STATION > 0)`. To have the testbed use the default gain model. 

On your third comment, I think we should still have it throw an error message. That's how it's done for the noise too and the user may not notice that the noise/gain models were set to default unless they read the print file.

Also, before I forget, is there a proper way to deal with comments here on GitHub? I'm doing the changes on my remote branch which I will push once I'm done

Settings.cc
AlanSalcedo: I couldn't think of any yet. These need knowledge of how many livetime settings are valid, which I don't see a way to anticipate.

marcomuzio: Fair enough. I suppose if one forgot to update this piece of code for a station the user will get another warning message that the requested gain file couldn't be found anyway. So I think it's okay to leave this as is. Thanks!

Detector.cc
marcomuzio: Yes absolutely agree regarding still throwing the error message.

Detector.cc
```diff
@@ -4154,6 +4156,7 @@ This function has two main parts: (1) loading of gain/phase values from gainFile
 	bool gainFileExists = (stat(filename.c_str(), &buffer)==0);
 	if (!gainFileExists){
 		sprintf(errorMessage, "Gain model file is not found (gain file exists %d) ", gainFileExists);
+		cout << "The gain filename is: " << filename << endl; 
 		throw std::runtime_error(errorMessage);
```
marcomuzio: I think this is where we want to switch to the default gain model right? Otherwise if the user inputs a valid station but not a valid config/no config the code will crash. I think this can be accomplished simply by doing something like:
`cerr << errorMessage << endl;`
`cerr << "Defaulting to in standard response" << endl;`
`filename = "./data/gain/custom_electronics.csv"`

Report.cc
marcomuzio: Bumping this @clark2668

clark2668: Apologies for my delay. I think Alan is right. `DETECTOR==4` means "load data like mode". Once you have selected `DETECTOR==4`, you need to pick `DETECTOR_STATION`, which sets the ARA station being emulated.

Re: the shift-indexing, I _think_ it will break in other places if we propagate that change (unfortunately).

marcomuzio: Okay, thanks for setting me straight there :) Then this all looks good to me as is. Thanks again to you both!

PR #64, opened 2023-04-03
new options that can customize energy spectrum range

MyoungchulK:
In continuous energy mode (`EXPENONT` < 5), the energy spectrum range is fixed to 10^16 ~ 21.5 eV (Based on the code, I believe the true intention was 10^16 ~ 22 eV...). 
But the calculation limit of the cross-section in `GetSigma()` in the Primary class is 10^16 ~ 21 eV based on _Connolly et al. 2011_
Because of this, the simulation skips the event if Nu energy happens to be above 10^21 eV (check Brian's [PR #28](https://github.com/ara-software/AraSim/pull/29))
This PR allows users to freely customize (avoid the skip or focus certain energy range) the energy spectrum range by `EXPONENT_MAX` and `EXPONENT_MIN` options. 
`EXPONENT_MIN`: lower limit for energy spectrum range. The default is 10^16 [eV]
`EXPONENT_MAX`: upper limit for energy spectrum range. The default is 10^22 [eV]
Let me know If you have a comment!


Settings.h
```diff
@@ -51,6 +51,10 @@ class Settings
 
         double EXPONENT; // 10^19 eV neutrinos only
 
+        double EXPONENT_MIN; // minimum eV of Neutrino energy spectrum, if user use continuous energy spectrum for simulating Neutrino
```
clark2668: Don't you mean minimum log10(eV)?

clark2668 (commented):
Small comment about docs.

Settings.h
MyoungchulK: indeed log10(eV)!

clark2668: Thanks!

clark2668:
Looks good Myounghcul, go ahead and resolve conflicts (you'll want to **keep both**, so we don't lose any comments), and then squash and merge.

PR #71, opened 2023-05-19
Re-calculate Nu position from antenna center

MyoungchulK:
We are only saving the Nu position (x, y, z, r, theta, phi) from the earth's center point of view. 
So, the true theta and phi values of Neutrino are not much useful for the current reconstruction method.
And in order to compare the reconstructed Nu position with the true Nu position, we (at least me) need to recalculate the true position from the antenna center point of view.
I decided to add another Nu position that calculated from the antenna center

`PosNuFromAntennaCenter()` in the `Primaries` class will load antenna and Nu positions, and apply simply built-in `SetXYZ()` in `Vector` class
The re-calculated value will be saved in the `posnu_from_antcen` array

Let me know if this calculation is already in AraSim!

### DIFF
--- a/Primaries.cc
+++ b/Primaries.cc
@@ -975,6 +975,10 @@ Interaction::Interaction(IceModel *antarctica, Detector *detector, Settings *set
     
     
   }
+
+  //! re-calculate Nu position (x, y, z, r, theta, phi) from antenna center point of view. MK added -2023-05-19-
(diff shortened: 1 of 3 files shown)


MyoungchulK:
the check failed at the installing ROOT!

clark2668:
I think the test must have run when CERN's servers were down. I've re-started the test above.

PR #74, opened 2023-05-26
Added missing include statement in Report header.

marcomuzio:
`Report` class uses structures from `Position` class but for some reason the include statement for `Position.h` is missing. This causes compilation errors for users if they `#include "Report.h"` without also including `#include "Position.h"`. 

I asked whether there was a reason for this missing include statement on #software-arasim on Slack, but no one had an explanation (or opposition to fixing this).

### DIFF
--- a/Report.h
+++ b/Report.h
@@ -9,6 +9,8 @@
 #include <vector>
 #include "TGraph.h"
 
+#include "Position.h"
+
 #ifndef __CINT__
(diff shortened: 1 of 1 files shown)


clark2668:
I think this is uncontroversial, you can just go for it. In fact, in the future, for a change this small, you can probably just commit directly to main.

marcomuzio:
Sounds good, thanks Brian! :)

PR #75, opened 2023-06-01
Adding trigger delays and trigger masking

AlanSalcedo:
This is pull request is an upgrade of trigger formation. It allows the user to add trigger delays and channel masking.


Detector.cc
```diff
@@ -4085,7 +4108,204 @@ This function has two main parts: (1) loading of gain/phase values from gainFile
...
+	//First, let's check that the file exists and the format is correct
+	
+	// check if the file with delays + masking exists
+	char errorMessage[400];
+        struct stat buffer;
+
+        bool trigFileExists = (stat(filename.c_str(), &buffer)==0);
+        if (!trigFileExists){
+		sprintf(errorMessage, "Trigger formation file is not found (trigger file exists %d) ", trigFileExists);
+        	throw std::runtime_error(errorMessage);
```
marcomuzio: Maybe this should just be an error message and then it will fall back to the default? Otherwise I think the code will always crash for A4 & A5 right now?

Detector.cc
```diff
@@ -4085,7 +4108,204 @@ This function has two main parts: (1) loading of gain/phase values from gainFile
...
+
+	ifstream trigFile(filename.c_str()); // open the file
+	string line; // a dummy variable we can stream over
+
+	// if the trigger formation file exists, then make sure our user has formatted the file correctly
+	// in particular, it means we really need to see the word "Channel No." as the first word in the header file 
+
+	string expected_first_column_header = "Channel";
+        if(trigFile.is_open()){
+                while(trigFile.good()){
```
marcomuzio: is this while loop necessary? if im reading it right it seems like it should either throw an error or break the loop on the first iteration? maybe it would be more clear what the code is doing if we remove the loop if it is superfluous

Detector.cc
```diff
@@ -4085,7 +4108,204 @@ This function has two main parts: (1) loading of gain/phase values from gainFile
...
+	// we have already verified that the file exists, so no need to check again...
+	trigFile.clear();
+        trigFile.seekg(0, ios::beg);
+
+
+	// Figure out how many columns we have, to check we get the four expected columns 
+	int numCommas = 0;
+        int theLineNo = 0;
+        if(trigFile.is_open()){
+                while(trigFile.good()){
```
marcomuzio: similarly here i think we dont need this loop? only the first iteration will actually do anything right?

Detector.cc
```diff
@@ -4085,7 +4108,204 @@ This function has two main parts: (1) loading of gain/phase values from gainFile
...
+                                                    chBin, temp_delay_val);
+                                           cout << "The trigger formation filename is: " << filename << endl;
+                                           throw std::runtime_error(errorMessage);
+                                }
+
+                                triggerDelay[chBin] = temp_delay_val;
+
+				//get and store the value for trigger masking decision
+				getline(trigFile, line, ',');
+                                double temp_mask_val = atof(line.c_str());
```
marcomuzio: maybe this should be an integer? i guess it shouldn't matter, but just to be safe we don't miss a 0 or 1 value due to some floating point representation.

Detector.cc
```diff
@@ -4085,7 +4108,204 @@ This function has two main parts: (1) loading of gain/phase values from gainFile
...
+                                }
+
+                                triggerMask[chBin] = temp_mask_val;	
+
+		
+				// once more to get the final column, this time we need to detect the newline character		
+			
+				//get and store the value to activate delay
+				getline(trigFile, line, '\n');
+                                double temp_activeDelay_val = atof(line.c_str());
```
marcomuzio: Similarly, I think we could make this an int (or a bool? but maybe that would be more error prone)

Detector.cc
```diff
@@ -4842,7 +5062,30 @@ void Detector::ReadRayleigh_New(Settings *settings1) {    // will return gain (d
...
+		offset = 0;
+	}
+	
+	return offset;
+}
+
+int Detector::GetTrigMasking( int ch){ 
 
+	int masking;
+	masking = triggerMask[ch];
```
marcomuzio: since you're not checking for errors or anything with this value, maybe you should just return it directly? i.e. `return triggerMask[ch];` and not declare `masking` at all

Report.cc
```diff
@@ -2178,8 +2178,21 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
...
+ 			    }
+
+
+			    if( detector->GetTrigMasking(channel_num-1)==1){ //Triggr Masking
+				trig_j++;
+				continue;	
+			    }
+
+			    int offset=0;
+			    offset = detector->GetTrigOffset(channel_num-1, settings1);
```
marcomuzio: is there a reason not to assign this at initialization? maybe it would be slightly more readable that way? up to you

marcomuzio:
Hey @AlanSalgo1 this looks really good to me on the whole. I noted a few places where I think the code could be slimmed down a bit/made more readable. But I don't see any major issues. 

One question I do have regarding the files that get read-in though: these have been generated for A1-3, are the folks in charge of A4 & A5 expected to generate the files for those stations?

AlanSalcedo:
Thanks, @marcomuzio! I agree with your comments and have addressed all of them. Let me know if I missed anything.

About the files, I think you're right. If the analyzers can tell me what livetime configs need cable delays taken into account as well as what channels need masking, I can add them as well. 

For A1, I knew the channels to be masked but not the livetime configs to add delays. I added masking but not delays. The delays_masking.csv files reflect that.

Detector.cc
marcomuzio: I think the update is probably fine as is, but just to let you know you can do `char` to `int` conversion directly using `atoi`

marcomuzio:
This all looks good to me @AlanSalgo1, I think you can go ahead and merge. Let's just make sure we have a plan for producing those delays_masking.csv files for the rest of the stations. We should discuss in the 5-station call tomorrow.

PR #78, opened 2023-06-21
Fix Divide by 0 issues in GetFresnel() and reflection angle switch

abigailbishop:
1) For reflection coefficients, make calculations with reflection angle, not launch angle. 
2) Use a limit for reflection coefficients when reflection angle goes to zero. Math attached in pdf. 
3) Use a limit for magnification factor when launch angle of a ray goes to zero. Math attached in pdf. 

[Divide by 0 Math.pdf](https://github.com/ara-software/AraSim/files/11826453/Divide.by.0.Math.pdf)


IceModel.cc
```diff
@@ -1848,21 +1841,31 @@ void IceModel::GetFresnel (
     double pol_perp_trg=0, pol_parallel_trg=0;
 
     // check if ray is reflected or not
-    if (refl_angle < PI/2.) {   // the ray is reflected at the surface
+    if (launch_angle < PI/2.) {   // the ray is reflected at the surface
```
AlanSalcedo: Is there a difference with the previous condition (`if (refl_angle < PI/2.)`) or are there equivalent?

IceModel.cc
```diff
@@ -1883,6 +1886,68 @@ void IceModel::GetFresnel (
...
+        shiftedraysolver->Solve_Ray( // solve shifted ray from neutrino to shifted antenna
+            posnu, posant_shifted, 
+            icemodel, shifted_ray_output, settings1, RayStep
+        );  
+    }
+
+    // If there is a ray solution for the shifted antenna, 
+    //   calculate the magnification factor with the full equation. 
+    //   Otherwise, use the ratio of index_of_refractions approximation
+    if ( shifted_ray_output.size() > 0 ) {
```
AlanSalcedo, line 1927: (L1926) I believe Solve_Ray() populates `shifted_ray_output` with 4 vectors. And the actual number of ray solutions is the size of any one of these vectors. Something like `shifted_ray_output[0].size()`

Probably worth checking if this makes any difference. Same for the previous in (L1915)

IceModel.cc
```diff
@@ -1807,21 +1808,13 @@ void IceModel::GetFresnel_pokey (double i_ang, double n1, double n2, double &r,
...
-        double refl_angle, Position &posnu, Vector &launch_vector, Vector &rec_vector, Settings *settings1, double &fresnel, double &mag,
+        double refl_angle, Position &posnu, Vector &launch_vector, Vector &rec_vector, Settings *settings1, double &fresnel,
         Vector &Pol // will read the polarization at the source and return polarization at the target antenna
         ) {
 
     double n1 = 1.35;   // index of refraction at the firn
     double n2 = 1.;     // index of refraction at the air
 
-    // calculate the magnification factor for plane / spherical wave case
-    if (settings1->WAVE_TYPE == 0) { // plane wave
```
AlanSalcedo, line 1818: (L1818) This case separation between plane vs spherical waves was removed (?)

abigailbishop: NuRadioMC doesn't distinguish between spherical and plane waves, but I put back spherical and plane wave handling when the NuRadioMC equation doesn't work and we need to use our approximation

IceModel.cc
abigailbishop: You're right but this is just checking that a ray path exists (since the array would be empty otherwise) so I think its fine. This phrasing is also consistent with AraSim phrasing elsewhere. Though it would be important to rephrase this the way you suggest if we were iterating the number of valid ray paths

IceModel.cc
abigailbishop: I'm not sure why I switched them, I was probably playing with something and forgot to revert a commit. But this is resolved now

PR #79, opened 2023-06-21
Refresh Event Generator and Event List mode

abigailbishop:
Major Changes: 
1) Added event list writing function (settings1->EVENT_GENERATION_MODE == 2)
2) Energy of events input in eV, not as log(eV) or this way: 400 + log10(settings1->PNU[inu_thrown])*10.


Event.cc
```diff
@@ -39,8 +39,8 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
         pnu = spectra1->GetNuEnergy();
         
         if (settings1->EVENT_GENERATION_MODE == 1){
-            // pnu = pow(10., settings1->PNU[inu_thrown]);
-            settings1->EXPONENT = settings1->PNU[inu_thrown];
+            pnu = settings1->PNU[inu_thrown];
+            settings1->EXPONENT = 400 + log10(settings1->PNU[inu_thrown])*10.;
```
marcomuzio: Is this change backwards-compatible? Maybe it's not a big concern at this point, but might be worth thinking about.

marcomuzio: Hey @abigailbishop this looks good to me overall, I don't have any major concerns. Could you update the `log.txt` to briefly summarize the changes made in this pull request?

marcomuzio:
Hey @abigailbishop this looks good to me overall, I don't have any major concerns. Could you update the `log.txt` to briefly summarize the changes made in this pull request?

Event.cc
abigailbishop: Oh good point I suppose it wouldn't be but I'm also not sure how much we've used event list mode in the past?

abigailbishop: I think I'll set an if then clause that says if the pnu is less than 1000 then the input pnu is read as an exponent. But if its between 1e6 and 1e13 then its read as energy in GeV and if its greater than 1e16 then its read as energy in eV. Then it'll be backwards compatible.

marcomuzio: Awesome, that sounds like a good solution. Probably you're right it won't effect anyone really but always better to be backwards-compatible if we can be. Thanks so much!

abigailbishop:
Thanks for the comments Marco! I made the energy handling back-compatible and added to the log file so I've finished addressing your comments.

PR #81, opened 2023-06-30
Add Phased Array Detector

abigailbishop:
Note: We might want to merge in the event generator pull request first. I'm not sure how to undo that many-commits-ago branch merger so the event generator changes are present in this pull request. 

Usage: 
    To build the phased array, use DETECTOR=5. 
    To use the Phased Array trigger functions in Report.cc, use TRIG_SCAN_MODE=5. 
    DETECTOR_STATION:
        1: Jan 2018 - Dec 2018 ( PA channels only)
        2: October 2019 - Jan 2020 (All PA channels and 1 A5 channel)
        3: Jan 2020 - present (PA + 7 VPols from A5)

Change summary: 

**AraSim.cc** 
    Loads in data used to estimate phased array trigger if TRIG_SCAN_MODE==5. 

**Detector.cc:** 
    Builds Phased Array and extra attached antennas using hard coded coordinates. 
    SetupInstalledStations() needs settings1 now so settings1->DETECTOR_STATION can be used to attach the appropriate number of ARA05 antennas to the Phased Array DAQ. 

**Primaries.cc:** 
    Told the PickNear functions to shift antennas for DETECTOR=5 so the detector isn't built away from where neutrinos are generated. 

**Report.cc:** 
    I made CircularBuffer belong to the report class and not triggerCheckLoop() so It could be used in checkPATrigger() too. 
    Added V_noise to Report::Antenna object for use in checkPATrigger(). 
   Differentiated window and global passing in triggerCheckLoop() because it affected PA trigger at one point though it doesn't anymore, but it is more semantically accurate with the window_pass_bit. 
    GetAverageSNR(), GetAverageSNR2(), IsTrigger(), and CheckPATrigger() are PA triggering functions I took from Kaeli's simulation 
         I've only made small edits to these like adding in voltage+noise workflow to GetAverageSNR2()
        Kaeli had GetAverageSNR() in use which takes the topmost PA Vpol, calculates the maximum SNR it sees from an events, and calls that the average SNR. That is currently the function still in use for triggering. 
        To trigger: the "average" SNR is interpolated in a trigger efficiency vs SNR plot Kaeli took from an old paper. If a random number (0-1, float) is less than the efficiency at the present SNR, the station triggers. 

**Settings.cc:** 
    Added error messages to catch anticipated user error/oversights when running PA simulations. 
    Added to comments in header file as well.


Detector.cc
```diff
@@ -1,6308 +1,6717 @@
...
-            
-            
-            
-            if ( params.antenna_orientation == 0 ) {    // all borehole antennas facing same x
-                stations[StationIndex].strings[stringNum].antennas[antennaNum].orient = 0;
-            }
-            else if ( params.antenna_orientation == 1 ) {   // borehole antennas one next facing different way
-                if ( stringNum==0||stringNum==3 ) {
-                    if ( antennaNum==0||antennaNum==1 ) {
-                        stations[Sta
```
AlanSalcedo: Can you run PA (mode=5) without problems? It should complain about this gain model. Or default to the .csv file in `./data/gain`  

This should be changed to `./data/gain/PA_Electronics_TotalGainPhase.csv`. You can check that's being loaded in the output file. The path is printed there.

AlanSalcedo: (L2098) @marcomuzio do we want separate custom gain file for the PA? There's a custom one for A1-5. 

@abigailbishop this should be either `./data/gain/custom_electronics.csv` or whatever custom file is added for the PA

Settings.cc
```diff
@@ -852,6 +852,21 @@ int Settings::CheckCompatibilitiesSettings() {
         num_err++;
     }
 
+    if (TRIG_SCAN_MODE==5 && ( DETECTOR!=5 && DETECTOR!=9 ) ){
```
AlanSalcedo: Why DETECTOR!=9?

Detector.cc
AlanSalcedo, line 1764: (L1764) This was added by Brian in mode=4. @marcomuzio should we have something similar in mode=5 (for PA only)? Also Marco, I see there's 5 noise files in your PR, why is that? @abigailbishop has 3 configs for the PA. How do we make this compatible?

Settings.cc
AlanSalcedo, line 1: It doesn't allow me to add comments outside of the green range. In ~L1019, we have the compatibilities for DETECTOR_LIVETIME_CONFIG. I think we need to make it so that it accepts values 1 and 2. 

You'll find L1019 inside an if(DETECTOR==4) statement. I think we'll need a separate one for DETECTOR==5 that allows DETECTOR_STATION between 1-3 only?

Detector.cc
AlanSalcedo, line 6310: (L6300) Do we need another for DETECTOR == 5 like you had in `GetChannelfromStringAntenna`?

AlanSalcedo: (L5987) Shouldn't this be DETECTOR_STATION==2?

AlanSalcedo: (L5968) I don't understand the numbering in this whole section :( before the if-statements

Report.cc
```diff
@@ -3262,22 +3259,30 @@ int Report::triggerCheckLoop(Settings *settings1, Detector *detector, Event *eve
...
 	 (settings1->TRIG_ONLY_LOW_CH_ON==1 && settings1->DETECTOR!=3 && antenna_i<2 ) ){ // channel filter: choose if to use lower/borehole channels or not
       
-
-      int channel_num = detector->GetChannelfromStringAntenna ( i, string_i, antenna_i, settings1 );
+      int channel_num;
+      if (detector->Get_mode()==5){
+          channel_num = detector->GetChannelfromStringAntenna ( 5, string_i, antenna_i, settings1 );
+      }
+      else {
+          channel_num = detector->GetChannelfromStringAntenna ( i, string_i, antenna_i, settings1 );
```
AlanSalcedo: I think this will crash for DETECTOR == 0-3 because of their channel numbering. In `GetChannelfromStringAntenna` (the one with settings as extra argument), you can see numbering shifted with +1 for DETECTOR mode 4 and 5. Only God knows why. 

I usually make an extra case for DETECTOR 0-3 where channel_num gets a +1

AlanSalcedo (commented):
@abigailbishop I finished going through it. I was okay reading most of it, except Report.cc. Don't know enough to comment there. 

@marcomuzio you'll probably want to check out my comments and see if I didn't suggest wrong things.

Detector.cc
marcomuzio: Hmm, I'm not sure. What's the difference between a file like `./data/gain/Electronics_TotalGainPhase.csv` and `./data/gain/custom_electronics.csv`?

marcomuzio: Sorry, what is the change Brian made that you're referring to?

I think there's 2 different kinds of configurations for PA. There are sort of detector configurations, which have to do with the channels plugged into the DAQ, and then there's livetime configurations, which have to do with the noise/gain models. As I understand, Abby has made it so that `DETECTOR_STATION` controls the detector configuration so I think that leaves `DETECTOR_LIVETIME_CONFIG` free for the livetime configuration? (Sorry if I misunderstand the question)

Settings.cc
abigailbishop: Because I forgot to remove it. That was Kaeli's DETECTOR id when she made the phased array, thanks for catching that!

Detector.cc
AlanSalcedo: Sorry for not being entirely clear. This is where per-channel noise files are loaded. So we need to copy this to mode 5 for the PA.

Also, since DETECTOR_STATION means different configs of the detector (I supposethe PA config changed in time), then a given livetime_configs only works with specific detector configs?

AlanSalcedo: Honestly, no clue. I guess it's just used for custom electronics so that people don't touch `Electronics_TotalGainPhase.csv`. I don't think we'll use it for ARA5 analysis, I just though of it for consistency.

Settings.cc
abigailbishop: I wansn't able to find DETECTOR_LIVETIME_CONFIG but I found DETECTOR_STATION_LIVETIME_CONFIG and told Settings.cc to complain if DETECTOR=5 and the values are less than 1 or greater than 2. Is it actually DETECTOR_LIVETIME_CONFIG that I need to modify? 

I have a DETECTOR_STATION catch already, is this what you're hoping for (it's around line 866)?
```
    if ( DETECTOR==5 && ( DETECTOR_STATION<1 || DETECTOR_STATION>3 ) ) {
        cerr<<"DETECTOR_STATION must be 1,2,3 for PA Station (DETECTOR = 5) "<<endl;
        num_err++;
    }
```

Detector.cc
abigailbishop: Alas it does get confused but I never noticed it because AraSim switches to a default gain model. I'll update this

```
     Reading standard ARA electronics response
The not found gain filename is: ./data/ARA_Electronics_TotalGain_TwoFilters.txt
Defaulted to standard response from file: ./data/gain/ARA_Electronics_TotalGain_TwoFilters.csv
```

Settings.cc
AlanSalcedo: Oops, they are separate issues :( 

The catch I'm talking about is for DETECTOR = 4, which then has subcases for DETECTOR_STATION. We are missing the case for DETECTOR_STATION = 5 (simulate A5 only?) which has 2 DETECTOR_STATION_LIVETIME_CONFIG values (1 or 2) according to Marco's data in `AraSim/data/gain`

abigailbishop: Good to go. For the DETECTOR_STATION_LIVETIME_CONFIG I've added a check for DETECTOR=5 before going into all the DETECTOR_STATION checks
```
        if(DETECTOR_STATION_LIVETIME_CONFIG>-1){

            if ( ( (int)DETECTOR_STATION == 5 ) && 
                 ( ( DETECTOR_STATION_LIVETIME_CONFIG>5 || DETECTOR_STATION_LIVETIME_CONFIG<1 ) ){
                    // Phased Array mode (DETECTOR=5) uses DETECTOR_STATION for DAQ configurations
                    // Need to check this before checking DETECTOR_STATION in the context of ARA stations
                    cerr<<" DETECTOR_STATION_LIVETIME_CONFIG is set to "<<DETECTOR_STATION_LIVETIME_CONFIG<<" but there are only five expected configurations for A1"<<endl;
                    num_err++;
            }
	    else if((int)DETECTOR_STATION==1){
```

And regarding the valid DETECTOR_STATION values I've already got this excerpt in Settings.cc 
```
    if ( DETECTOR==5 && ( DETECTOR_STATION<1 || DETECTOR_STATION>3 ) ) {
        cerr<<"DETECTOR_STATION must be 1,2,3 for PA Station (DETECTOR = 5) "<<endl;
        num_err++;
    }
```

I just want to check, does this address both of your comments?

Detector.cc
marcomuzio: Wait so `Electronics_TotalGainPhase.csv` is the default gain model, right? And then `./data/gain/custom_electronics.csv` is like if you wanna simulate some theoretical gain for example? If that's the case then I could see it being useful to have the capability to do that for the PA separately. But probably for now it can just be a copy of `./data/gain/custom_electronics.csv`?

abigailbishop: Alright, is this where you noticed this? 
```
        // read total elec. chain response file!!
        cout << "start read elect chain" << endl;
        if (settings1 -> CUSTOM_ELECTRONICS == 0) {
            cout << "     Reading standard PA electronics response" << endl;
            ReadElectChain("./data/gain/PA_Electronics_TotalGainPhase.csv", settings1); // Here's the change I made
        } else if (settings1 -> CUSTOM_ELECTRONICS == 1) {
            //read a custom user defined electronics gain
            cout << "     Reading custom electronics response" << endl;
            ReadElectChain("./data/custom_electronics.txt", settings1);
        }
        cout << "done read elect chain" << endl;
```

clark2668: The custom electronics file is frankly a vestige. I don't even think it's formatted correctly...?

abigailbishop: Not sure if this applies here as well but does this change address this comment? 

```
        // read total elec. chain response file!!
        cout << "start read elect chain" << endl;
        if (settings1 -> CUSTOM_ELECTRONICS == 0) {
            cout << "     Reading standard PA electronics response" << endl;
            ReadElectChain("./data/gain/PA_Electronics_TotalGainPhase.csv", settings1);  // Here's the change I made
        } else if (settings1 -> CUSTOM_ELECTRONICS == 1) {
            //read a custom user defined electronics gain
            cout << "     Reading custom electronics response" << endl;
            ReadElectChain("./data/custom_electronics.txt", settings1);
        }
        cout << "done read elect chain" << endl;
```

marcomuzio: Ah yes, that makes sense. I think we will need to add that for PA. I'm not sure how @abigailbishop is pulling in the noise/gain models now for the time being?

The gain/noise models shouldn't cause any errors if they are used with the wrong detector configuration I don't think. But in reality some livetime configurations only existed during one of the detector configurations. But maybe we can leave getting that right to the user?

AlanSalcedo: Yeah, you're right. So we can make `PA_custom_electronics.csv` being just a copy of `custom_electronics.csv` and load it here

AlanSalcedo: I think so, it's L2093

AlanSalcedo: > The custom electronics file is frankly a vestige. I don't even think it's formatted correctly...?

@clark2668 I made it a csv and reformatted it x_x

abigailbishop: Great thank you :)

AlanSalcedo: Alright! @abigailbishop we need a block like L1759 - L1772, copied to mode = 5 loading files `sigmavsfreq_PA_config_{livetime_config}.csv`

@marcomuzio I didn't see data-driven noise for mode 5 (PA) yet

Settings.cc
AlanSalcedo: Yes. Thanks! And Marco just said somewhere else, it's fine to have DETECTOR = 5 take DETECTOR_STATION_LIVETIME_CONFIG = (1-5) and not worry about their compatibility with DETECTOR_STATION

abigailbishop: Okay great thank you!

marcomuzio: Sorry I just want to double check one thing: the DETECTOR_STATION values only specify PA configurations right? None of them are just A5? Just want to make sure were pulling in the A5 gain/noise models when we simulate that.

Detector.cc
abigailbishop: While we're talking about pulling stuff in from DETECTOR=4, @marcomuzio do you think I need the following Calpulser code pulled in as well? 

```
        // if calpulser case
        if (settings1 -> CALPULSER_ON > 0) {
            // read TestBed Calpulser waveform measured (before pulser)
            ReadCalPulserWF("./data/CalPulserWF.txt", settings1);
        }
```

marcomuzio: I'm not sure? This is to simulate a calpulser I guess? I imagine it would be useful. Annoying its in a particular DETECTOR section

abigailbishop: Maybe its only in sections where single stations are simulated instead of arrays? (In which case it may behave oddly in the DETECTOR=5 mode since you can make a hexagonal array of phased array stations though it's set to 1 station right now (that's how Kaeli had it) )

Settings.cc
AlanSalcedo: Right. 

For A5 we have `DETECTOR = 4`, `DETECTOR_STATION = 5` and we're calling noise/gain files how it was done already:

`./data/noise/sigmavsfreq_A_%d_config_%d.csv` and `./data/gain/In_situ_Electronics_A%d_C%d.csv`

For the PA we have `DETECTOR = 5`, `DETECTOR_STATION` specifies DAQ configs, and we're using only `DETECTOR_STATION_LIVETIME_CONFIG` to point to noise files:

`./data/noise/sigmavsfreq_PA_config_%d.csv`

The gain for PA is only one file. So that hardcoded: `PA_Electronics_TotalGainPhase.csv`

Detector.cc
marcomuzio: Ya that could be, maybe you can have it throw an error if that function is called and there's more than 1 PA station? Or at least a warning message.

Settings.cc
marcomuzio: Okay great, thanks for confirming Alan! Closing this one out.

Detector.cc
abigailbishop: Yep you're totally right, thanks again :)

abigailbishop: This was an annoying hack that I really should add extra commenting to and get the approval of like @marcomuzio for. 

But from what I experienced, the InstalledStations object saves channel numbers for the purpose of getting various antenna-related models so I told the PA Vpols to all obey the modeling for RF Channel 5 (string 2 bottom pol) and all the PA Hpols to obey the modeling for RF channel 13 (string 2 bottom hpol) since I didn't know if we had any per-channel modeling for the PA. I may be toooootally wrong in all of this though.

abigailbishop: Oh shoot maybe its string 0 actually (still bottom H and Vpols), since the channel numbers in AraSim don't match the chart from the bootcamp slides (like we chatted about on Thursday)

abigailbishop: I can add it but I haven't yet discovered it being an issue (not to say its not actually an issue). I added the DETECTOR==5 bit to GetChannelfromStringAntenna because AraSim yelled at me, I haven't been yelled at about Detector::GetSSAfromChannel or any of the other functions in that area of code yet

marcomuzio: I don't think that there are any PA-specific antenna models. I think they are just the same antennas so it makes sense. Not sure if for the ARA stations those models even vary beyond being a Vpol vs a Hpol. I think this should be okay!

AlanSalcedo: I am still concerned, not about antenna models but by how AraSim -> RF channel mapping is done. 

My suspicion is that AraSim uses these numbers for mapping, since `GetChannelfromStringAntenna` checks these values. 

That would cause mapping several antennas to the same RF channel during noise modelling. And noise models are different per channel for the PA?

AlanSalcedo: A quick check would be to add these lines I borrowed from mode 4 into mode 5 (after setting up the station) from the Detector constructor :

        for (int j = 0; j < stations[0].strings.size(); j++) {
            for (int k = 0; k < stations[0].strings[j].antennas.size(); k++) {

                cout <<
                    "DetectorStation2:string:antenna:X:Y:Z:chno :: " <<
                    j << " : " <<
                    k << " : " <<
                    stations[0].strings[j].antennas[k].GetX() << " : " <<
                    stations[0].strings[j].antennas[k].GetY() << " : " <<
                    stations[0].strings[j].antennas[k].GetZ() << " : \t" <<
                    GetChannelfromStringAntenna(stationID, j, k, settings1) <<
                    endl;

                params.number_of_antennas++;
            }
        }
        
We can take a look at the mapping an make sure each antenna is mapping to different RF channels

abigailbishop: Great thanks Marco and Alan, I've added the ReadRayleighFit_DeepStation and calpulser codes  from DETECTOR=4 (with a warning if the calpulser is run with more than one station being simulated)

marcomuzio: It looks like the else statement has `InstalledStations[1]` instead of `InstalledStations[stationId]` for DETECTOR==4? Does `InstalledStations[1]` make sense for DETECTOR==5?

abigailbishop: Sorry, what's left to be done in this thread? 

Do I just need to take `ReadElectChain("./data/custom_electronics.txt", settings1);` and change it to `ReadElectChain("./data/PA_custom_electronics.txt", settings1);`? 

Do I also need to make that CSV file or did you already do that Alan?

AlanSalcedo: I think you need `ReadElectChain("./data/gain/PA_custom_electronics.csv", settings1);` and make that file a copy of `custom_electronics.csv`. It is in `./data/gain` already

marcomuzio: > Sorry, what's left to be done in this thread?
> 
> Do I just need to take `ReadElectChain("./data/custom_electronics.txt", settings1);` and change it to `ReadElectChain("./data/custom_electronics.txt", settings1);`?
> 
> Do I also need to make that CSV file or did you already do that Alan?

Change to what?

abigailbishop: Alright, that's been done. Thank you1

abigailbishop: I'm building the DETECTOR==5 for this right now, but having duplicates in InstalledStations.VHChannel makes GetSSAfromChannel potentially inaccurate in a way that's associated with the "I don't understand the numbering in this section" so I'm going to switch to that thread

abigailbishop: Upon further inspection I think the channel numbering in this section needs unique numbers for each channel so I'll go back to having the PA channels  use a channel number > 100 and wherever the code was looking for antenna models via InstalledStations, I'll add in a catch to force PA vpols and hpols to use the antenna models I was trying to make them use with this hack

AlanSalcedo: So in reality PA antennas aren't they RF channels numbered 0-15?  

 I used `GetChannelfromStringAntenna` to point to columns in the noise files so we kinda need that numbering

abigailbishop: Oh you're totally right, I don't remember where I got it in my head that phased array channel numbers are over 100, it may have been another relic from Kaeli's version of this code or something I saw in a presentation somewhere. Good grief. I'm updating these now to do 0-15 channel numbering

Report.cc
abigailbishop: ugh this did happen! I set the highest channel number to 15 and AraSim broke saying "Hi I have no record of channel number 16". Bonkers. I think I have to do the thing I've seen in Report.cc where I tell AraSim to subtract 1 from whatever GetChannelfromStringAntenna  returns when its figuring out gain models

abigailbishop: Alright there are two areas in Report.cc where the +1 is countered with a -1 for DETECTOR=4 so I added DETECTOR=5 to that if-then statement as well. I just ran DETECTOR=1 and 2 and both worked fine so I think this is good to go

Detector.cc
abigailbishop: Alright I have resolved my whacky numbering haha. I made the channel numbering match my strings in DETECTOR=5 and not the channel numbers in InstalledStation[1-5] (where we get confused about which strings are which). 

If you're all not worried about that, we can close this thread.

abigailbishop: I added it, thanks :)

abigailbishop: Actually, I take that back. Every so often I go back and forth on how channels are numbered and I'm oscillating again. I'll message when I sort it out

AraSim.cc
```diff
@@ -403,6 +403,66 @@ int main(int argc, char **argv) {   // read setup.txt file
     int inu = 0;
     int Events_Thrown = 0;
     int Events_Passed = 0;
+
+    // Prepare arrays required by Phased Array sim
+    double snr_PA[60];
```
marcomuzio: maybe this is a dumb question, but where do these array sizes come from? is there a variable we can use instead so that they aren't "magic numbers"?

AraSim.cc
```diff
@@ -403,6 +403,66 @@ int main(int argc, char **argv) {   // read setup.txt file
...
+        infile.open("nuphase_trig_effc.txt",ios::in);
+        if(infile.fail()) // checks to see if file opended
+            {
+            cout << "   error, file could not be opened" << endl;
+            return 1; // no point continuing if the file didn't open...
+            }
+            int num = 0;
+            while(!infile.eof()) // reads file to end of *file*, not line
+            {
+                infile >> snr_PA[num] && // read first column number
```
marcomuzio: Let's rewrite these lines in the following way for clarity: 
`infile >> snr_PA[num] // read first column number`
`       >> eff_PA[num]; // read second column number`
 `++num;`

AraSim.cc
```diff
@@ -403,6 +403,66 @@ int main(int argc, char **argv) {   // read setup.txt file
...
+        infile.open("nuphase_SNR_angle.txt",ios::in);
+        if(infile.fail()) // checks to see if file opended
+        {
+            cout << "   error, file could not be opened" << endl;
+            return 1; // no point continuing if the file didn't open...
+        }
+        num = 0;
+            while(!infile.eof()) // reads file to end of *file*, not line
+            {
+                infile >> angle_PA[num] &&  // read first column number
```
marcomuzio: Same thing here

Report.h
```diff
@@ -237,11 +263,22 @@ class Report {
     
 //    void Connect_Interaction_Detector (Event *event, Detector *detector, RaySolver *raysolver, Signal *signal, IceModel *icemodel, Settings *settings1, Trigger *trigger);
     
-    void Connect_Interaction_Detector_V2 (Event *event, Detector *detector, RaySolver *raysolver, Signal *signal, IceModel *icemodel, Settings *settings1, Trigger *trigger, int evt);    
+    void Connect_Interaction_Detector_V2 (Event *event, Detector *detector, RaySolver *raysolver, Signal *signal, IceModel *icemodel, Settings *settings1, Trigger *trigger, int evt, double* xdata, double* ydata, double* ang_data, double* snr_data);
```
marcomuzio: How widely is this function used? Is adding these extra arguments going to break some folks code?

AraSim.cc
```diff
@@ -403,6 +403,66 @@ int main(int argc, char **argv) {   // read setup.txt file
...
+            {
+                infile >> snr_PA[num] && // read first column number
+                infile >> eff_PA[num] &&// read second column number
+                ++num;
+                // you can also do it on the same line like this:
+                // infile >> exam1[num] >> exam2[num] >> exam3[num]; ++num;
+            }
+            cout<<"   number of triggering efficiency data points: "<<num<<endl;
+        infile.close();
+        // std::cout << "The data from Phased array" << '\n';
```
marcomuzio: If we can, I think it would be good to clean up blocks of commented code that aren't useful anymore.

Detector.cc
```diff
@@ -1,6308 +1,6760 @@
...
```
marcomuzio: What happens in this case? Will the code keep running. Maybe we should throw an error? I would also print this message with `cerr` rather than `cout`

marcomuzio, line 1987: Do other stations hard code these? I wonder if we should be reading them in?

Detector.h
```diff
@@ -505,7 +505,7 @@ class Detector {
     vector < IdealStation > IdealStations;
     
 
-    void SetupInstalledStations();
+    void SetupInstalledStations(Settings *settings1);
```
marcomuzio, line 508: is `settings1` actually used? do we need to add this argument?

Report.cc
```diff
@@ -278,6 +279,113 @@ void Antenna_r::clear_useless(Settings *settings1) {   // to reduce the size of
...
+    if(mode>1) last_value=buffer[i];// in mode 1 we don't care about the values, just about the addToNPass
+        
+    if(input_value<pthresh){
+        addToNPass++; // if value entering buffer is over threshold we increase the counter.
+        buffer[i]=input_value;
+    }
+    else buffer[i]=0; // if under threshold just insert zero
+      
+    // improve best threshold value in the buffer
+    if(mode>1&&buffer[i]<best_value){ best_value=buffer[i]; temp_value=best_value; changelog=1;}
```
marcomuzio: Nitpicking but can we break this out into a few lines?

marcomuzio: and add some spaces in the if condition haha

Report.cc
```diff
@@ -278,6 +279,113 @@ void Antenna_r::clear_useless(Settings *settings1) {   // to reduce the size of
...
+    i++;
+    if(i==N) i=0;
+
+    return changelog;
+
+}// fill
+    
+double CircularBuffer::findBestValue(){
+    double temp_best=0;
+    for(int ii=0;ii<N;ii++) if(buffer[ii]<temp_best) {
```
marcomuzio: lets break this into a few lines and add some braces too:
`for(int ii=0;ii<N;ii++) {`
`  if(buffer[ii]<temp_best)`
`    temp_best=buffer[ii];`
`}`

Report.cc
```diff
@@ -278,6 +279,113 @@ void Antenna_r::clear_useless(Settings *settings1) {   // to reduce the size of
...
+int CircularBuffer::numBinsToOldestTrigger(){
+    
+    if(addToNPass<1){ 
+        cerr<<"ERROR: there are no triggers in this buffer right now!\n"; 
+        return -1;
+    }
+
+    int j; // the number of bins after "i" where the first trigger is found 
+    for(j=1; j<N; j++){
+        // if(i==0&&buffer[N-1]<0) return 1;
```
marcomuzio: Let's remove these commented lines of code where we can

Report.cc
```diff
@@ -1810,6 +1922,7 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
                                         else
                                         {
                                             // if we are checking previous string, check upto entire antennas
+                                            if (settings1->DETECTOR==5 && settings1->DETECTOR_STATION==3 && j_sub==3 && k_sub==1) continue; // Avoids segfault from string 3 having 1 antenna for PA DETECTOR_STATION 3
```
marcomuzio: lets put this `continue` on another line so the lines not too long

Report.cc
```diff
@@ -3377,9 +3378,8 @@ int Report::triggerCheckLoop(Settings *settings1, Detector *detector, Event *eve
 //       if(scan_mode==1) return last_trig_bin; //  if we aren't going to scan all the Pthresh values, just return
       if(scan_mode==1) return trig_i; //  if we aren't going to scan all the Pthresh values, just return
     }
-    else global_pass_bit=0;// no trigger
     
-    if(scan_mode>1&&check_TDR_configuration&&global_pass_bit){// if there's a trigger and anything changes in the buffers, restock the TDR arrays
+    if(scan_mode>1&&check_TDR_configuration&&window_pass_bit){// if there's a trigger and anything changes in the buffers, restock the TDR arrays
```
marcomuzio: add some spaces in this condition

Report.cc
```diff
@@ -3978,7 +3978,7 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
 
 
 // this one is for single signal
-void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger, Detector *detector, int signalbin, vector <double> &V, int *noise_ID, int ID, int StationIndex) {
+void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger, Detector *detector, int signalbin, vector <double> &V, int *noise_ID, int ID, int StationIndex, vector <double> *V_with_noise) {
```
marcomuzio, line 3993: Probably this is only used in this file? But does adding this argument break things elsewhere?

Report.cc
```diff
@@ -4040,7 +4044,7 @@ void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger,
 
 
 // this one is for two connected signals 
-void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger, Detector *detector, int signalbin1, int signalbin2, vector <double> &V1, vector <double> &V2, int *noise_ID, int ID, int StationIndex) {
+void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger, Detector *detector, int signalbin1, int signalbin2, vector <double> &V1, vector <double> &V2, int *noise_ID, int ID, int StationIndex, vector <double> *V_with_noise) {
```
marcomuzio, line 4059: same as above

Report.cc
```diff
@@ -4118,7 +4125,7 @@ void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger,
 
 
 // this one is for three connected signals 
-void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger, Detector *detector, int signalbin0, int signalbin1, int signalbin2, vector <double> &V0, vector <double> &V1, vector <double> &V2, int *noise_ID, int ID, int StationIndex) {
+void Report::Select_Wave_Convlv_Exchange(Settings *settings1, Trigger *trigger, Detector *detector, int signalbin0, int signalbin1, int signalbin2, vector <double> &V0, vector <double> &V1, vector <double> &V2, int *noise_ID, int ID, int StationIndex, vector <double> *V_with_noise) {
```
marcomuzio, line 4140: omg why do we have so many versions of this function?? lol

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
   return getHitTimesVector(detector, station_i, 1);
   
 }
+
+double Report::getAverageSNR2(int raysolnum, int station_i, int trig_analysis_mode){
+    double total_snr = 0.0;
+    double temp_snr = 0.0;
+    double peak;
+    for (int ant_num = 2; ant_num<9; ant_num++){
```
marcomuzio: why does this start from `ant_num = 2`?

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+
+double Report::getAverageSNR2(int raysolnum, int station_i, int trig_analysis_mode){
+    double total_snr = 0.0;
+    double temp_snr = 0.0;
+    double peak;
+    for (int ant_num = 2; ant_num<9; ant_num++){
+        peak = 0.0;
+        
+        if(trig_analysis_mode == 2) { // Noise only triggers
+            peak=3.5*0.04;
```
marcomuzio: do we know what these magic numbers are? where do they come from?

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+            }
+        }
+
+        temp_snr = peak/0.04;
+        if(temp_snr>25) temp_snr = 25; // Cap SNR at 25
+        total_snr = total_snr+temp_snr;
+
+    }
+
+    total_snr = total_snr/7.0;
```
marcomuzio: instead of 7 can we have a counter or a variable like `num_ant`?

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+
+                    }
+                }
+            }
+            else {
+                peak = 0.0;
+            }
+        }
+
+        temp_snr = peak/0.04;
```
marcomuzio: where does this 0.04 come from?

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+){
+    // Calculates max SNR in topmost PA Vpol 
+    //   and multplies it by viewing angle factors. 
+    // Then determines signal efficiency by interpolating oberved SNR from 
+    //   efficiency vs SNR data.
+    // Triggers if signal efficiency is above certain treshold.
+    // If triggered, saves relevant information.
+
+    //cout <<"successfully made it to PA Trigger!" << endl;
+    int BINSIZE = 1200/(settings1->TIMESTEP*1.e9);
```
marcomuzio: what is 1200?

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+    while(raySolNum < stations[i].strings[0].antennas[8].SignalBin.size()){
+
+        int signalbinPA = stations[i].strings[0].antennas[8].SignalBin[raySolNum]; //new kah
+        int bin_value;
+        double noise_rms = 0.04; //The noise RMS for an ARA waveform
+        double avgSnr;
+        if(settings1->TRIG_ANALYSIS_MODE == 2) { // Noise only triggers
+            avgSnr=3.5;
+        }
+        // I don't trust the signal+noise trigger estimator right now - ARB 7/7/23
```
marcomuzio: Is this something we need to address?

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+                int last_trig_bin = signalbinPA;
+                int my_ch_id = 0;
+                stations[i].Global_Pass = last_trig_bin;
+                for (size_t str = 0; str < detector->stations[i].strings.size(); str++) {
+                    for (size_t ant = 0; ant < detector->stations[i].strings[str].antennas.size(); ant++) {
+                        double peakvalue = 0;
+                        for (int bin=0; bin<BINSIZE; bin++) {
+
+                            bin_value = signalbinPA - BINSIZE/2 + bin;
+                            // stations[i].strings[str].antennas[ant].V_mimic.push_back( ( trigger->Full_window_V[ant][ last_trig_bin - settings1->NFOUR/4 + mimicbin ] )*1.e3 );// save in mV
```
marcomuzio: lets trim out this block of commented code if its not needed

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+
+            double *TDR_all_sorted_temp;
+            double *TDR_Vpol_sorted_temp;
+            double *TDR_Hpol_sorted_temp;
+            
+            if(settings1->TRIG_SCAN_MODE>1){ // prepare TDR storage arrays 
+                TDR_all_sorted_temp=new double[numChan];
+                TDR_Vpol_sorted_temp=new double[numChanVpol];
+                TDR_Hpol_sorted_temp=new double[numChanHpol];
+                for(int trig_j=0;trig_j<numChan;    trig_j++) TDR_all_sorted_temp[trig_j] =0;
```
marcomuzio: lets add some spaces in this `for` loop condition and put the code in the for loop on a separate line

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+                
+                    int string_i = detector->getStringfromArbAntID( i, trig_j);
+                    int antenna_i = detector->getAntennafromArbAntID( i, trig_j);
+
+                    trig_window_bin = (int)(settings1->TRIG_WINDOW / settings1->TIMESTEP);  // coincidence window bin for trigger
+
+                    int channel_num = detector->GetChannelfromStringAntenna( 5, string_i, antenna_i, settings1 );
+
+                    // assign Pthresh a value 
+                    if(settings1->NOISE_CHANNEL_MODE==0) Pthresh_value[trig_j] = trigger->Full_window[trig_j][trig_i] / ( trigger->rmsdiode                   * detector->GetThresOffset( i, channel_num-1,settings1) );
```
marcomuzio: lets put the code inside the if statement on a separate line for readability

Report.cc
```diff
@@ -5679,3 +5691,484 @@ vector<double> Report::getHitTimesVectorHpol(Detector *detector, int station_i){
...
+
+                    // fill the buffers (if any changes occur mark check_TDR_configuration as non-zero)
+                    if(trig_i<trig_search_init+trig_window_bin) check_TDR_configuration+=buffer[trig_j]->fill(Pthresh_value[trig_j]);
+                    else check_TDR_configuration+=buffer[trig_j]->add(Pthresh_value[trig_j]);
+
+                    // if there is at least one value above threshold in the buffer, this is ++
+                    if( buffer[trig_j]->addToNPass>0 && 
+                        string_i == 0 && 
+                        detector->stations[i].strings[string_i].antennas[antenna_i].type == 0
+                    ) N_pass_V++;
```
marcomuzio: lets put the parenthesis on the previous line, i find this hard to read. sorry for all this nitpicking!

Settings.cc
```diff
@@ -1002,11 +1024,21 @@ int Settings::CheckCompatibilitiesSettings() {
...
+            else if((int)DETECTOR_STATION==5){
+                if(DETECTOR_STATION_LIVETIME_CONFIG>2 || DETECTOR_STATION_LIVETIME_CONFIG<1){
+                    cerr<<" DETECTOR_STATION_LIVETIME_CONFIG is set to "<<DETECTOR_STATION_LIVETIME_CONFIG<<" but there are only two expected configurations for A3"<<endl;
+                    num_err++;
+                }
+            }
             else{
                 cerr<<" DETECTOR_STATION_LIVETIME_CONFIG is set to "<<DETECTOR_STATION_LIVETIME_CONFIG<<" but DETECTOR_STATION is "<<DETECTOR_STATION<<endl;
-                cerr<<" DETECTOR_STATION_LIVETIME_CONFIG is only valid for A2 and A3 "<<endl;
+                cerr<<" This is not currently permitted in via error handling in Settings.cc "<<endl;
```
marcomuzio: this error message is a little confusing, can you rephrase?

Report.cc
abigailbishop: We have 3 versions I think. The first for when we process 1 wave through antennas, the second when 2 are processed, and the third when 3 are processed.

Report.cc
abigailbishop: Because antenna 0 and antenna 1 are the bottom HPols on the PA and they aren’t used for triggering

Report.cc
abigailbishop: Which magic numbers? 3.5 is what Kaeli had as a typical Noise SNR but I should track down where it comes from and add a reference in the code

Report.cc
abigailbishop: Addressing it would involve changing the trigger pipeline from 
“Base the trigger on a random number compared to the signal efficiency of the PA (based on the max SNR and how on/off cone the signal is)
To 
“Actually create the antenna beams out of the waveforms and determine triggering like the PA board does”

I’m not solid on the mechanics of making beams so it’d take me a little bit to put that together but I can if you’d like. Either Kaeli’s paper or the PA design paper said the “random number vs pa efficiency” method is accurate to within 15% iirc

Detector.h
abigailbishop: Yeah I need `settings1->DETECTOR_STATION` to build the correct number of A5 antennas into the PA board.

abigailbishop: When `DETECTOR_STATION` is 1 there are no A5 antennas connected, when it’s 2 there is 1 A5 antenna, and when it’s 3 there are 7. I’m fairly sure `DETECTOR_STATION` isn’t stored anywhere in the detector class so I need to access it through `settings1`

marcomuzio: Okay that sounds good! We should just make sure adding this argument doesn't break anything elsewhere. Or have a default argument to a `nullptr` if it's not provided maybe?

Report.cc
marcomuzio: Right that and 0.04. I would just add a comment saying what it is or store it in a descriptively name variable eg `noiseSNR`

Report.cc
marcomuzio: Okay that makes sense, can you just add a comment making that clear?

Report.cc
marcomuzio: Okay ya I think we can just stick with that, maybe we can just add a comment that we're going to use the pa efficiency instead and this is work for the future.

Report.cc
marcomuzio: Lordy lordy, well nothing to do about that in this PR. Thanks!

AraSim.cc
abigailbishop: Done

AraSim.cc
abigailbishop: Done

Detector.cc
abigailbishop: I'm guessing you're talking about this line and related DETECTOR==5 part of the function? https://github.com/abigailbishop/AraSim/blob/742170fa7422b3c13ea4c6e6d0c0f25ffb4c6559/Detector.cc#L6264
This function could use a freshening up because I think each DETECTOR=__ part is just for setting `stationId`. So what was going wrong originally is that `stationId` wasn't being set properly for DETECTOR=5 meaning the wrong channel numbers would be returned (and usually it would run for a bit before breaking) or the script wouldn't find a channel number and would return `-1` which makes the code seg fault quickly.

abigailbishop: Can you send a line number?  I'm not sure what you're referring to here

marcomuzio: (Sorry maybe these comments only show the context properly when you look at the full file diffs)

This is referring to line 1869 (new), specifically the "else" case where the input station is invalid.

Detector.h
abigailbishop: I'm okay with that but do we need to make anullptr? Every time AraSim is run it initializes a settings object: 
https://github.com/ara-software/AraSim/blob/969fbf0ae8b7c5184584c832aee98b9b639ce955/AraSim.cc#L46

 Unless you're expecting someone to run just a specific function in AraSim without running the whole thing?

Detector.cc
marcomuzio: This is referring to line 1986 (new) where the first channel depth is set

Detector.h
marcomuzio: I just suggest that in case this function is used somewhere without the new settings argument (maybe you've checked if `SetupInstalledStations()` is called anywhere else in the code. Having a default argument would allow for this, but now that I think about it probably the code wouldn't compile without the default argument in the case that `SetupInstalledStations()` is called. 

nullptr was suggested just becuz that's the only default argument that works for pointers unless you have some global pointer (in which case you wouldn't need to pass it to the function).

abigailbishop: Yeah that function is only called once (when the detector class is first initialized). But I set the default value to nullptr anyways.

Report.cc
abigailbishop: D1

abigailbishop: Whoops I just wanted to click "Comment" not "start review". Not sure what I just did

abigailbishop: I'm not even sure where to find that but we can resolve this once the "Pending" flag is somehow removed

Report.cc
abigailbishop: D1

abigailbishop: This one is also a pending review now too? This happens when I comment from the "file diff" screen so I will stop doing that haha

abigailbishop: This can also be resolved once we figure out how to remove the "Pending" flag from my comments here

Report.cc
abigailbishop: D1

Report.cc
abigailbishop: Yep d1

Report.cc
abigailbishop: d1

Report.cc
abigailbishop: d1

Report.cc
abigailbishop: I changed it to what I pasted below. The parentheses is still on the line below but I added other changes that may make this better? 

```
                    if ( 
                        buffer[trig_j]->addToNPass>0 && 
                        string_i == 0 && 
                        detector->stations[i].strings[string_i].antennas[antenna_i].type == 0
                    ) {
                        N_pass_V++;
                    }
```

Report.cc
abigailbishop: d1

Report.cc
abigailbishop: d1 (I also split the if statement into different lines and explained the different clauses in the if statement)

Report.cc
abigailbishop: d1

Report.cc
abigailbishop: d1

Report.cc
abigailbishop: d1

abigailbishop: Oo should I add this as an official "GitHub Issue" so we don't forget we want to do this?

Settings.cc
abigailbishop: It is now ` This settings configuration is not currently anticipated by error handling in Settings.cc` is that better?

AraSim.cc
abigailbishop: It's just the number of lines in the files that are read in to fill these arrays. Can I just add that as a comment so we don't have to allocate memory for two integers that are only used in these four lines for array declaration?

AraSim.cc
abigailbishop: d1

abigailbishop (commented):
Testing this to see if my comments show up after “finishing my review”

Report.cc
marcomuzio: we can! that would probably be a good way to keep track of it. maybe we can add it as a feature request since its not an issue exactly right now

Report.cc
marcomuzio: ya i think this is better, thanks!

Settings.cc
marcomuzio: Ah I see what you're trying to say. Maybe simply `Unknown DETECTOR_STATION!`?

AraSim.cc
marcomuzio: I think that's fine, so long as we are sure that the file won't change (idk what the file is so maybe that's obvious). Or if it does, we should just be sure that either this doesn't silently read only a part of the file or alternatively write it so that it can adapt to read-in any length file.

Report.cc
abigailbishop: (Unresolving because I forgot to double check the new argument doesn’t break anything like in the last two comments)

Report.h
abigailbishop: It's used everywhere but I tested other modes of the code that don't use the PA and they ran just fine. 

But I moved these arrays from AraSim.cc to Trigger.cc so this function no longer has these arguments so this is no longer an issue.

AraSim.cc
abigailbishop: The files won't change.  They're nuphase_SNR_angle.txt and nuphase_trig_effc.txt. Kaeli pulled the numbers from old studies (I got a reference for the latter, but I need to ask her about where the former came from) so they should be static. 

(Also I moved this code into Trigger.cc and Trigger.h)

Trigger.cc
```diff
@@ -83,6 +83,45 @@ Trigger::Trigger(Detector *detector, Settings *settings1) {
...
+        while(!infile.eof()) { // reads file to end of *file*, not line
+            infile >> snr_PA[num]  // read first column number
+                   >> eff_PA[num]; // read second column number
+            ++num;
+        }
+        cout<<"   number of triggering efficiency data points: "<<num<<endl;
+        infile.close();
+
+        //load Angle vs SNR curve:
+        // Unsure how this curve was created but shapes of each beam resemble
```
abigailbishop: Note to self: Ask Kaeli how this curve from nuphase_SNR_angle.txt was created

Detector.cc
abigailbishop: Yeah if this block of code is reached, AraSim won't run properly. So I did change this from `cout` to `cerr`. Is that all that needs to happen for an error to be thrown?

Report.cc
abigailbishop: It is only in this file and I double checked that I updated every instance of this function and the other two related to it. I don't expect it to break anything anywhere especially since the effective volume test runs properly with the update. Does that sound alright then?

Settings.cc
abigailbishop: d1

Detector.cc
abigailbishop: I think the other specific station simulations read in the depths from AraRoot since they use the araGeomTool. For ideal station simulations antenna depths are determined by variables in the setup file. 

Not having the depths hard coded does sound like the better thing to do. I can make a file in AraSim, if you want, with all the antenna depths and have that read in during the detector's initialization. I don't know how to add the station to AraRoot so I'd need help with that if that's what you'd prefer.

Report.cc
abigailbishop: d1

Detector.cc
marcomuzio: To throw an error (and cause things to crash in a controlled way) you should use `throw runtime_error("Useful error message");` (or at least that's what I do, idk if it's the _best_ practice)

Report.cc
marcomuzio: Sounds good to me, thanks!

Detector.cc
marcomuzio: Actually I think we should do this the same way as for other stations but calling on the relevant nuphaseroot functions. I can take care of this in another pull request and you can leave it hard-coded for now. (Except if there's a version of this for the A5-only simulation? then maybe it's easy enough to implement the AraRoot functions for that part?)

abigailbishop: Great thanks. Done

Report.cc
abigailbishop: Resolving along with previous comment of similar nature:

> Abby Bishop: It is only in this file and I double checked that I updated every instance of this function and the other two related to it. I don't expect it to break anything anywhere especially since the effective volume test runs properly with the update. Does that sound alright then?
> 
> Marco Muzio: Sounds good to me, thanks!

Report.cc
abigailbishop: Resolving along with previous comment of similar nature:

> Abby Bishop: It is only in this file and I double checked that I updated every instance of this function and the other two related to it. I don't expect it to break anything anywhere especially since the effective volume test runs properly with the update. Does that sound alright then?
> 
> Marco Muzio: Sounds good to me, thanks!

Detector.cc
abigailbishop: Okay I've fixed the channel numbers finally. Closing this out.

Report.cc
abigailbishop: Added a comment explaining its ARA rms used historically (Kaeli and I couldn't track down a source)

Report.cc
abigailbishop: Kaeli wasn't sure. I added a comment though.

Report.cc
abigailbishop: 3.5 was an SNR that would force events to trigger and was used for noise only events. I did make it a variable and add a comment (for 3.5 and 0.04)

PR #83, opened 2023-07-12
New trigger and masking / New files 

AlanSalcedo:
Apparently, delay_active = 1 means that the DAQ box activated delays already and hence the waveforms are 'in phase'. I previously understood delay_active = 1 as if the user wanted to activate delays in AraSim, but that's not right. If the DAQ box already did that, there's no need to add them again. 

Note: Trigger/masking files for A5 are placeholders right now. I'm still missing information. They are created to not apply any delays or masking for now.


marcomuzio:
Hi Alan, this all looks good to me. What's the plan to get the A5 data? (We can also discuss this at the WG meeting tomorrow.)

AlanSalcedo:
Hi @marcomuzio, hope ICRC is going well!

I think this is complete now. For A5, I looked at raw data that Paramita sent me for masked-ant values and noticed that channels 14 and 15 are masked after run ~3,500 and config 2 starts at ~4,500. So I decided to mask them in config 2 only.

marcomuzio:
All sounds good to me, thanks Alan!

PR #84, opened 2023-07-21
Add capability to simulate A1 as an ATRI or ICRR 

AlanSalcedo:
Prior to this change, A1 was simulated as an ICRR. **A1 is currently an ATRI!**

This is how we used to simulate it: DETECTOR_STATION = 1 (In AraSim settings), then this is used as StationID = ARA_STATION1 (in AraRoot) which describes A1 as an ICRR. 

Now we can also do: DETECTOR_STATION = 100 (In AraSim settings) -> StationID = ARA_STATION1B (in AraRoot) which corresponds to A1 as an ATRI.

Please note that this poses the following problem: AraSim only handles DETECTOR_STATION = 1 to build A1. To deal with this, you'll see a new setting (not able to be modified by the user) called DETECTOR_STATION_ARAROOT used in functions involving AraRoot tools. 

For DETECTOR_STATION = 100, we store this value in DETECTOR_STATION_ARAROOT = 100 and default DETECTOR_STATION back to 1. This way AraSim builds A1 normally and uses information retrieved from AraRoot (like AraGeomTool tables) using DETECTOR_STATION_ARAROOT. This only happens for DETECTOR_STATION = 100. For other values, DETECTOR_STATION_ARAROOT = DETECTOR_STATION keeping other stations "untouched". 

**In summary, you can now simulate A1 as ATRI with DETECTOR_STATION = 100 and as ICRR with DETECTOR_STATION = 1.**


Settings.cc
```diff
@@ -362,6 +363,11 @@ void Settings::ReadFile(string setupfile) {
               }
               else if (label == "DETECTOR_STATION") {
                   DETECTOR_STATION = atof( line.substr(line.find_first_of("=") + 1).c_str() );
+		  DETECTOR_STATION_ARAROOT = DETECTOR_STATION;
+		  if (DETECTOR_STATION == 100) {
+			DETECTOR_STATION = 1
+			DETECTOR_STATION_ARAROOT = 100;
```
AlanSalgo1, line 369: This line is probably unnecessary but I didn't see it at the time. I could delete it. It doesn't hurt either.

Detector.cc
```diff
@@ -5414,7 +5414,7 @@ void Detector::PrepareVectorsInstalled(int importedStation) {
 }
 
 
-void Detector::SetupInstalledStations() {
+void Detector::SetupInstalledStations(Settings *settings1) {
```
marcomuzio: Does adding this break things elsewhere?

Detector.cc
```diff
@@ -5482,42 +5482,87 @@ void Detector::SetupInstalledStations() {
...
 
-        InstalledStations[1].nSurfaces = InstalledStations[1].surfaceChannels.size();
+        	InstalledStations[1].nChannels = 20;
+        	InstalledStations[1].nChannelsVH = 16;
 
-        InstalledStations[1].nChannels = 20;
-        InstalledStations[1].nChannelsVH = 16;
+    	}
+	
+	else{ //Otherwise always build A1 as an ATRI
```
marcomuzio: do we need to worry about `settings1-> DETECTOR_STATION_ARAROOT` having some unreasonable value? I'm just wondering if a blanket `else` statement might let some errors sneak by silently?

marcomuzio:
Thanks for doing this @AlanSalgo1, looks like it took a lot of careful work keeping track of things! I just have a few small comments/questions but no major concerns.

marcomuzio:
It occurred to me during the analysis call, it would be good to include a print statement when the user is simulating A1 to call out if it's being simulated as an ICRR or ATRI explicitly.

AlanSalcedo:
> It occurred to me during the analysis call, it would be good to include a print statement when the user is simulating A1 to call out if it's being simulated as an ICRR or ATRI explicitly.

I just added this.

PR #87, opened 2023-10-04
Fix Various Bugs in Phased Array PR

abigailbishop:
Commit 8a93090: I had changed the RF `ch_limit` to 28 when I was trying to simulate both the PA and A5 DAQ in the same simulation. I thought I reverted all the 28s back to 16s but must've missed these two. If 28, AraSim will seg fault on a triggered event once `MakeUsefulEvent()` tries to access the 17th antenna (which doesn't exist). The variable was originally set to 16.

Commit 45f2710: In `GetNoiseWaveforms_ch` with `NOISE=1`, I forgot to add the PA Detector (DETECTOR=5) to a block of code that builds noise waveforms specially for `DETECTOR=3` and `DETECTOR=4` . As a result, when generating noise waveforms for PA channels individually, the following arrays wouldn't be populated (which caused seg faults when the code tried to access them): `Vfft_noise_after`, `Vfft_noise_before`. Also, the variable `vnoise` would not be updated. 

Commit 5c1d4fb: Another scenario where I didn't realize I needed to add `else if settings1->DETECTOR==5` to a block of code. This code block adjusts the variable `stationID` which affects where AraSim looks for channel numbering.

I have now searched AraSim for the (caps sensitive) phrase `DETECTOR == 4`, `DETECTOR==4`, `DETECTOR== 4`, and `DETECTOR ==4` in an attempt to find more of these scenarios, leading to commits a6768c4 and 7f2486d


Report.cc
```diff
@@ -4936,6 +4936,57 @@ void Report::GetNoiseWaveforms_ch(Settings * settings1, Detector * detector, dou
...
+                Vfft_noise_after.push_back(vnoise[2 * k + 1]);
+
+                // and apply the inverse FFT normalization factor
+                vnoise[2 * k] *= 2. / ((double) settings1 -> DATA_BIN_SIZE);
+                vnoise[2 * k + 1] *= 2. / ((double) settings1 -> DATA_BIN_SIZE);
+
+            }
+
+            // real FT back to get vnoise in time domain waveform; 
+            Tools::realft(vnoise, -1, settings1 -> DATA_BIN_SIZE);
```
marcomuzio, line 4987: Is this the function used for the Fourier transform for other stations to make the noise? Versus a FFTtools implementation? Not sure if there's any difference though

Settings.cc
```diff
@@ -1046,6 +1038,28 @@ int Settings::CheckCompatibilitiesSettings() {
...
+	    cerr << "DETECTOR is set to 5" << endl; 
+
+        // Phased Array mode (DETECTOR=5) uses DETECTOR_STATION for DAQ configurations
+	    if (DETECTOR_STATION < 1 || DETECTOR_STATION > 3) {
+	        cerr << "DETECTOR_STATION is not set to a valid station number" << endl;
+	        num_err++;
+	    }
+
+        if ( DETECTOR_STATION_LIVETIME_CONFIG > -1 ) {
+            if ( DETECTOR_STATION_LIVETIME_CONFIG>5 || DETECTOR_STATION_LIVETIME_CONFIG<1 ) {
```
marcomuzio, line 1053: Not sure what this is used for later down or if it's just for error checking, but is it possible to avoid hardcoding that the max `DETECTOR_STATION_LIVETIME_CONFIG` value is 5? Otherwise everytime a new config file is added, we need to remember to come update this

marcomuzio:
Thanks @abigailbishop! I just had two small comments, but otherwise I think this all looks good!

Settings.cc
abigailbishop: I think it’s mostly for error checking but it’s hard coded for A1-A5 (which is a poor excuse, I know). Maybe I could write something to query the directory with the livetime files to figure the limits out that way?

marcomuzio: Okay fair enough. Maybe let's not worry about it for now then. We can upgrade later if needed.

Report.cc
abigailbishop: I just copied and pasted what was written for DETECTOR=4 (so all the other stations). I’m not sure why I didn’t add DETECTOR=5 to the DETECTOR=4 block now that I think about it. Not sure how to answer your FFTtools question though

marcomuzio: I think it's all okay then, just seems better to be uniform in the approach between stations if we can. (Clearly I was just trying to find _something_ to comment on lolol)

abigailbishop: Sounds good!

PR #94, opened 2023-11-10
Birefringence

AlanSalcedo:
This update adds the capability to model birefringence in AraSim. 

We activate birefringence with BIREFRINGENCE=1. Then, we can model the uniaxial case with BIAXIAL=0 or the biaxial case with BIAXIAL=1.

An introduction is here: https://aradocs.wipac.wisc.edu/0029/002928/001/Birefringence%20Analysis%2011-06-23.pdf

In this implementation, I have made sure of a few important things:

1. Changes are minimal in preexisting classes like Report.cc. 

In birefringence, we have a single ray solution split into two rays which propagate acquiring a time difference and have separate perpendicular polarizations. The part that describes this in Report.cc is now included in a for-loop which loops twice for BIREFRINGENCE=1 and once for BIREFRINGENCE=0, as it should since there's only one ray without birefringence.

The large changes you see in Report.cc here happen because I added a tab for all lines inside the just mentioned loop. 

All major changes are in Birefringence.cc.

2. I have checked that output files are the same for AraSim-prior-to-this-update and for BIREFRINGENCE=0. i.e. When we turn off birefringence, we recover our previous framework.


AraSim.cc
```diff
@@ -25,6 +25,7 @@
 #include "Spectra.h"
 #include "Tools.h"
 #include "Trigger.h"
+#include "Birefringence.hh"
```
marcomuzio: Is there a reason for this to be a .hh file?

Settings.cc
```diff
@@ -347,6 +349,12 @@ void Settings::ReadFile(string setupfile) {
               else if (label == "NOFZ") {
                   NOFZ = atof( line.substr(line.find_first_of("=") + 1).c_str() );
               }
+	      else if (label == "BIREFRINGENCE"){
+		  BIREFRINGENCE = atof( line.substr(line.find_first_of("=") + 1).c_str() );
```
marcomuzio, line 353: I think in these settings you want to use `atoi` so that the string is directly converted to an integer. Unless there's a reason to do the implicit conversion from a float?

Birefringence.cc
```diff
@@ -0,0 +1,804 @@
...
+      		n1vec.push_back(thisn);
+     	}
+
+    	n2file >> stemp;
+    	for (int i=0;i<NDEPTHS_NS;i++) {//loops through this data
+      		n2file >> thisdepth >> thisn;//piping into thisn
+      		vdepths_n2.push_back(-1.*thisdepth);
+      		if (BIAXIAL==1)//
+			n2vec.push_back(thisn);//adds our data into thisn for certain properties
+      		else if (BIAXIAL==0 || BIAXIAL==-1)
```
marcomuzio, line 490: Is this currently supported or not? Your comment above made me think it isn't?

Birefringence.cc
```diff
@@ -0,0 +1,804 @@
...
+
+      		int thissize=(int)n1vec.size();//length of vector
+
+      		//empties the vectors
+      		n1vec.clear();
+      		n2vec.clear();
+      		n3vec.clear();
+      
+      		//adds the first values to the vectors
+      		for (int i=0;i<thissize;i++) {//why do we need a loop?
```
marcomuzio: Was this comment resolved? May cause confusion in the future?

Birefringence.cc
```diff
@@ -0,0 +1,804 @@
...
+  
+} // End of getDeltaN()
+
+
+void Birefringence::Read_Indicatrix_Par(string sn1file, string sn2file, string sn3file, Settings *settings1){ //reads in data from n1file, n2file, n3file. These define the principal axes of the indicatrix
+
+
+	int BIAXIAL = settings1->BIAXIAL; //BIAXIAL = 1 is the only allowed now (by settings inconsistencies) for modeling biaxial birefringence, but I'll leave the code and capability for uniaxial birefringence (BIAXIAL = 0) and isotropic (BIAXIAL = -1). -ASG 11/7/2023 
+
+	int NDEPTHS_NS=81; //POSSIBLE IMPROVEMENT: This is not agnostic to the number of rows (depths) in sn1file. We may want to determine this by counting the number of lines.
```
marcomuzio: Maybe for now we can add some sort of catch so that there's an error message or an error is thrown if there are a different number of lines?

Birefringence.hh
```diff
@@ -0,0 +1,58 @@
...
+	vector<double> vdepths_n2;
+	vector<double> vdepths_n3;
+	TVector3 p_e1;
+    	TVector3 p_e2;
+	TVector3 p_e1_src;
+	TVector3 p_e2_src;
+
+	int CONSTANTINDICATRIX=0; ///PLACEHOLDER! I don't think we'll ever need CONSTANTINDICATRIX = 1. Used in Birefringence::Read_Indicatrix_Par() inside Birefringence.cc
+	const double CLIGHT=3.E8;
+	const double HOWSMALLISTOOSMALL=1.e-8;
```
marcomuzio: lol this is great

Settings.cc
marcomuzio: idk maybe it doesn't matter. clearly this has been done a lot in this file...

marcomuzio:
Hi Alan, really nice job with this! I have no major concerns after a quick scan. I left a few small comments you can take as you like, but I do think we should probably change the .hh to a .h file? Maybe there's a good reason for it but I'd be worried the Makefile might miss that header.

Can you update the log file before merging?

AraSim.cc
AlanSalcedo: Nah, just changed to .h

Settings.cc
AlanSalcedo: Okay, I'll leave this

Birefringence.cc
AlanSalcedo: BIAXIAL=0 is supported. I rewrote my comment more clearly.

Birefringence.cc
AlanSalcedo: Yeah, I just deleted it

Birefringence.cc
AlanSalcedo: I decided to not be lazy and actually count them haha. It's fixed now

AlanSalcedo:
Hi @marcomuzio, I just addressed the comments above. It might be ready now

Birefringence.hh
AlanSalcedo: The `HOWSMALLISTOOSMALL` variable? Amy had other interesting naming like `findthefreakingaxis` in her code haha

AraSim.cc
marcomuzio: Great, thanks!

Birefringence.hh
marcomuzio: haha thats amazing

Birefringence.cc
marcomuzio: Awesome, thanks!

marcomuzio:
Thanks so much @AlanSalcedo, I think this is good to be committed. Nice work!

PR #98, opened 2024-01-02
PA Gain Models

marcomuzio:
Updated PA noise model to account for adc-to-voltage conversion. Added in-situ PA gain models and updated Detector class to use these when requested.


Detector.cc
```diff
@@ -2109,8 +2109,21 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
         cout << "start read elect chain" << endl;
         if (settings1 -> CUSTOM_ELECTRONICS == 0) {
             //read the standard ARA electronics
```
abigailbishop, line 2118: Is this comment now out of date?

Detector.cc
```diff
@@ -2109,8 +2109,21 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
...
+            cout << the_gain_filename <<endl;
+            struct stat buffer;
+            bool gainFileExists = (stat(the_gain_filename, &buffer)==0);
+            if(gainFileExists) {
+              ReadElectChain(std::string(the_gain_filename), settings1);
+            }
+            else {
+              cout <<"    In situ gain model does not exist for this station"<<endl;
+              cout << "     Reading standard PA electronics response" << endl;
+              ReadElectChain("./data/gain/PA_Electronics_TotalGainPhase.csv", settings1);  // Here's the change I made
```
abigailbishop: Should this also throw a warning or something? I'm not sure if people consistently read through the entire output from AraSim jobs and they may miss this? 

(Do we still need your comment "Here's the change I made" too?

Detector.cc
```diff
@@ -2109,8 +2109,21 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
...
+            cout <<" Reading in situ ARA electronics response for the PA and this configuration from file:"<<endl;	
+            cout << the_gain_filename <<endl;
+            struct stat buffer;
+            bool gainFileExists = (stat(the_gain_filename, &buffer)==0);
+            if(gainFileExists) {
+              ReadElectChain(std::string(the_gain_filename), settings1);
+            }
+            else {
+              cout <<"    In situ gain model does not exist for this station"<<endl;
+              cout << "     Reading standard PA electronics response" << endl;
```
abigailbishop: This is so nitpicky, sorry, can you remove one space from this string right before "Reading standard..." so it matches the number of spaces in "In situ gain model.." (Unless you intended this)

abigailbishop:
Right now, this is looking good to me! I was able to successfully run this and trigger the PA without seg faults. 

Have you done any quick verifications of waveforms or effective volumes before and after this change to make sure they're as you expect? The veff_check that GitHub runs automatically doesn't use the PA detector so it shouldn't provide any information on this pull request.

Detector.cc
marcomuzio: I guess these are still standard as opposed to the "custom" electronics files?

Detector.cc
marcomuzio: Done!

Detector.cc
marcomuzio: We could print the messages about using the standard response into error stream? Do you think that would work?

marcomuzio:
Ah no, I haven't done any of that yet. Can we just update the Aeff estimate first after the merge? (Should we add a PA Veff check to the automatic script?)

Detector.cc
clark2668: I would recommend we throw a fatal warning -- that is, crash out. We don't want this to silently do something the user doesn't expect.

marcomuzio: We had explicitly tried to avoid that before so that the user could still do something if the noise/gain model wasn't made yet. But maybe the perspective is different now? I can update it for the normal stations as well.

clark2668: I mean, I would have said that if someone wants to simulate a yet unconfigured setting, they can always copy the default file and change the name. But that way what they're doing is active (choosing to fill the file with "dummy" values), instead of AraSim doing it for them. But that's just my two cents.

marcomuzio: Okay, fair enough. I'll implement some crashes.

Detector.cc
abigailbishop: Oh yeah I guess 0 = False in code speak. No worries then

Detector.cc
abigailbishop: Thanks!

abigailbishop:
I'm pro adding an effective volume check that checks PA effective volume. I wasn't trying to ask you to run the full Aeff estimate right now haha, just wanted to check that the effective volume didn't change too much before and after your change

marcomuzio:
Ah good point! Okay I've added this to the list of feature requests, hopefully we can get to that soon.

Detector.cc
marcomuzio: Okay added! Let me know if you think this looks good. @AlanSalcedo can you take a quick look at the normal ARA station errors I added? Hopefully it works.

AlanSalcedo: That works. Though it would probably be better to do that here: https://github.com/ara-software/AraSim/blob/9674d4b0f87cd7d3a539021a13f33c589432fc89/Detector.cc#L4148

That's where it's done for noise modeling, it saves you doing it for each detector mode and extra if-else statements.

marcomuzio: Ah, thanks! I moved the runtime errors to that function so I think it should be good now.

PR #110, opened 2024-03-13
Update Ant H_eff and Restructure Report.cc

FlahertyJustinC:
Updating the antenna effective height calculation.  Calculation was double-counting two factors: a factor of 1/2 that corrected an issue with a previous definition of the effective height, and a factor of 1/sqrt(2) that accounted for a 3 dB splitter in the electronics chain.  That 3 dB splitter only applied to A1-3, whereas A4-5 followed a different setup.  This split in power is now folded into the electronics gain files.

### DIFF
--- a/Report.cc
+++ b/Report.cc
@@ -4760,7 +4760,7 @@ void Report::ApplyElect_Tdomain_FirstTwo(double freq0, double freq1, Detector *d
 }
 
 void Report::ApplySplitterFactor(double &vm_real, double &vm_img, Settings *settings1) {
-    // Apply splitter/attenuation factor based on station.
+    // Apply splitter/attenuation factor in the digitizer path based on station.
     // AraSim default had factor of 1/sqrt(2), as does A1-3.  A4-5 have either a 1.4 dB or 10 dB factor, which is being checked.
     // See talk by Brian discussing splitter factors: https://aradocs.wipac.wisc.edu/cgi-bin/DocDB/ShowDocument?docid=2751
     
--- a/Report.cc
+++ b/Report.cc
@@ -4774,18 +4774,18 @@ void Report::ApplySplitterFactor(double &vm_real, double &vm_img, Settings *sett
         vm_real *= 1/sqrt(pow(10,-0.14));
         vm_img *= 1/sqrt(pow(10,-0.14));
     }
-    //Case for Phased Array implementation by Abby Bishop.  Will use placeholders of 1/sqrt(2) for now.
+    //Case for Phased Array implementation by Abby Bishop.  Will use placeholders of 1 for now.
     else if (settings1->DETECTOR == 5 and settings1->DETECTOR_STATION == 1) {
-        vm_real *= 1/sqrt(2);
-        vm_img *= 1/sqrt(2);    
+        vm_real *= 1;
+        vm_img *= 1;    
     }
     else if (settings1->DETECTOR == 5 and settings1->DETECTOR_STATION == 2) {
-        vm_real *= 1/sqrt(2);
-        vm_img *= 1/sqrt(2);    
+        vm_real *= 1;
+        vm_img *= 1;    
     }
     else if (settings1->DETECTOR == 5 and settings1->DETECTOR_STATION == 3) {
-        vm_real *= 1/sqrt(2);
-        vm_img *= 1/sqrt(2);    
+        vm_real *= 1;
+        vm_img *= 1;    
     }    
     //Final case where if none of the above are satisfied, it defaults to the historical 1/sqrt(2)
     else {
--- a/Report.cc
+++ b/Report.cc
@@ -1,4 +1,4 @@
-t#include "Detector.h"
+#include "Detector.h"
 #include "Report.h"
 #include "Event.h"
 #include "RaySolver.h"
@@ -66,7 +66,7 @@ void Report::delete_all() {
     noise_phase.clear();    // random noise phase generated in GetNoisePhase()
     signal_bin.clear();      // the center of bin where signal should locate
     signal_dbin.clear();     // the bin difference between signal bins
-    tt_signals.clear();    // if ray_sol time delay is small enough to connect each other
+    connect_signals.clear();    // if ray_sol time delay is small enough to connect each other
     Passed_chs.clear();
 
 
--- a/Report.cc
+++ b/Report.cc
@@ -1,4 +1,4 @@
-#include "Detector.h"
+t#include "Detector.h"
 #include "Report.h"
 #include "Event.h"
 #include "RaySolver.h"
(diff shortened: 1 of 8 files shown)


FlahertyJustinC:
Just realized I need to update the Veff test with the new effective height calculation.  Let me address that and push it.

FlahertyJustinC:
Veff test has been updated to reflect the new effective height calculation.  We now get four passing events in the Veff test.  Please check my work to make sure I didn't flub something.

marcomuzio:
Hi Justin, thanks for fixing this! Looks good to me overall. My only hesitation to saying we should merge it is the deleting of the sqrt(2) factors for the power splitters. Is it obvious where we need to replace these factors later? Maybe we can keep those in for now so we don't miss these when we replace is them with a getSplitterFactor kind of function.

On a similar note, can you correct the log file to omit the comment that these factors are folded into the electronics gain? Thanks!

FlahertyJustinC:
> Hi Justin, thanks for fixing this! Looks good to me overall. My only hesitation to saying we should merge it is the deleting of the sqrt(2) factors for the power splitters. Is it obvious where we need to replace these factors later? Maybe we can keep those in for now so we don't miss these when we replace is them with a getSplitterFactor kind of function.
> 
> On a similar note, can you correct the log file to omit the comment that these factors are folded into the electronics gain? Thanks!

If these factors aren't folded into the gain files, it would be an easy flag that can be added to the electronics gain step that applies the factor based on the station.  I'll look into it.

Report.cc
```diff
@@ -5311,17 +4740,58 @@ void Report::ApplyElect_Tdomain(double freq, Detector *detector, double &vm_real
...
 
     vm_bin0 = vm_bin0 * detector->GetElectGain_1D_OutZero( freq0 , gain_ch_no);
     vm_bin1 = vm_bin1 * detector->GetElectGain_1D_OutZero( freq1 , gain_ch_no);
 
+    //Apply power splitter/attenuator based on station.
+    GetSplitterFactor(vm_bin0, vm_bin1, settings1);
+
+}
+
+void Report::GetSplitterFactor(double &vm_real, double &vm_img, Settings *settings1) {
```
marcomuzio: Small thing, but maybe given this implementation it's better to call this function `ApplySplitterFactor`? Also we should make it explicit somewhere this is meant to be the splitter factor for the digitizer path.

SETUP/setup_TxSimulationVariablePulserDepth.txt
```diff
@@ -74,7 +78,7 @@ RAYSOL_RANGE=5000 // (DEFAULT=5000) distance from detector above which we do not
  
 USE_INSTALLED_TRIGGER_SETTINGS=0 // (DEFAULT=0) 0: ideal detector trigger settings. 1: use triggering as implemented in real stations (testbed only!). 
  
-TRIG_ANALYSIS_MODE=0 // (DEFAULT=0) 0: signal+noise. 1: signal only. 2: noise only 
+TRIG_ANALYSIS_MODE=1 // (DEFAULT=0) 0: signal+noise. 1: signal only. 2: noise only
```
marcomuzio, line 81: This changes the default right? Is this what we want?

marcomuzio: Oh maybe this is for a specific simulation?

Report.cc
```diff
@@ -5311,17 +4740,58 @@ void Report::ApplyElect_Tdomain(double freq, Detector *detector, double &vm_real
...
+    // Apply splitter/attenuation factor based on station.
+    // AraSim default had factor of 1/sqrt(2), as does A1-3.  A4-5 have either a 1.4 dB or 10 dB factor, which is being checked.
+    // See talk by Brian discussing splitter factors: https://aradocs.wipac.wisc.edu/cgi-bin/DocDB/ShowDocument?docid=2751
+    
+    //Case for simulating real ARA Stations 0 (testbed), A1, A2, and A3.
+    if (settings1->DETECTOR == 4 and settings1->DETECTOR_STATION < 4) {
```
marcomuzio, line 4768: Based on the part you found I think we can change this to `1/sqrt(pow(10,-0.34))`

Report.cc
```diff
@@ -5311,17 +4740,58 @@ void Report::ApplyElect_Tdomain(double freq, Detector *detector, double &vm_real
...
+        vm_real *= 1/sqrt(2);
+        vm_img *= 1/sqrt(2);
+    }
+    //Case for simulating real ARA Stations A4 and A5 (non phased array).
+    else if (settings1->DETECTOR == 4 and settings1->DETECTOR_STATION > 4) {
+        vm_real *= 1/sqrt(pow(10,-0.14));
+        vm_img *= 1/sqrt(pow(10,-0.14));
+    }
+    //Case for Phased Array implementation by Abby Bishop.  Will use placeholders of 1/sqrt(2) for now.
```
marcomuzio: The PA has no splitter, so I think these all are just `1`. Splitter is on DAQ (right?), so for the merged system even the A5 channels will have no splitter.

SETUP/setup_TxSimulationVariablePulserDepth.txt
FlahertyJustinC: Yeah this is just for a general setup file for Tx simulations.

Report.cc
FlahertyJustinC: For A2?  The part I showed was in the A2 DAQ.  I can see what's in the A4 DAQ as it's here at OSU (I think).

FlahertyJustinC: According to Patrick, A4 and A5 have a variable attenuator.  So I think this depends on the station config, and he implied that Brian may have folded that into the electronics gain for those stations.  We can take this to Slack to ask him.

Report.cc
FlahertyJustinC: Updated and pushed the name change, as well as added a comment in the function to state it's part of the digitizer path.

Report.cc
marcomuzio: Sorry, ya I meant for A1-3. Maybe Patrick is thinking of the calpulser attenuator? Seems a bit strange to have a variable splitter for the trigger/digitizer, but I'm no expert.

SETUP/setup_TxSimulationVariablePulserDepth.txt
marcomuzio: Got it, sorry for the spam!

Report.cc
marcomuzio: Thanks!

Report.cc
FlahertyJustinC: Okie doke I'll change it for A1-3 and we can revisit A4-5 later.

Report.cc
FlahertyJustinC: I can change those to ones if you want.  I wasn't sure, so I just kept the status quo and wrote it in a way that makes things easier for Abby.

marcomuzio: Ya we can implement now. I'm pretty sure that's right and it's easy fix later if needed.

Report.cc
marcomuzio: Thanks!

PR #113, opened 2024-03-20
Pulser sims

FlahertyJustinC:
Updated formatting of gain files. There were discrepancies in the whitespace between values that caused issues with the atof() function used in the Detector class functions ReadVgain(), ReadVgainTop(), ReadHgain(), and ReadTxgain(). This led to some minus signs getting truncated, causing a positive gain to be applied and saturating the waveform. This didn't appear until a recent commit where I changed the function to import the dB gain rather than the linear gain, so this should have no affect on past analyses.


marcomuzio:
If I understand correctly, this is just changing the number of spaces in these files? So assuming no numerical values were accidentally modified, I think we can just commit. Though if we're worried this could introduce any errors, we could just skip this commit and merge my open PR? I think that will take care of this problem anyway.

FlahertyJustinC:
> If I understand correctly, this is just changing the number of spaces in these files? So assuming no numerical values were accidentally modified, I think we can just commit. Though if we're worried this could introduce any errors, we could just skip this commit and merge my open PR? I think that will take care of this problem anyway.

Yes it simply just changed the numbers.  Attached is the short python code I used to reformat it, in case you want to check the logic.  It simply read in a line, split it into the numbers, then applied the spaces accordingly.

[reformatGain.py.txt](https://github.com/ara-software/AraSim/files/14783596/reformatGain.py.txt)

abigailbishop:
What checks did you do to make sure that all these files were read in correctly? I know its extra, but did you run simulations using each of these files to ensure they're being read in properly now?

abigailbishop:
Do you mind sharing the variable in AraSim you used to diagnose this issue too so that if we ever need to revisit this, the next person looking into this knows where to start looking?

marcomuzio:
I think I would lean towards our just scrapping this PR at this point, since the PR that was merged today should fix it anyway.

FlahertyJustinC:
> I think I would lean towards our just scrapping this PR at this point, since the PR that was merged today should fix it anyway.

Works for me.  I'll go ahead and close this request.

PR #118, opened 2024-03-21
Updated antenna gain file read-in

marcomuzio:
Updated antenna gain file read-in to be robust to various numbers of spaces/tabs, so long as the delimiter is whitespace, and to do some basic error checking. Also unified all antenna gain file read-in into a single function and added enums to encode antenna type.

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -2240,10 +2240,10 @@ inline void Detector::ReadAllAntennaGains(Settings *settings1){
     }
     
     //Read in antenna gain files.
-    ReadVgain(VgainFile, settings1);
-    ReadVgainTop(VgainTopFile, settings1);
-    ReadHgain(HgainFile, settings1);
-    ReadTxgain(TxgainFile, settings1);    
+    ReadAntennaGain(VgainFile, settings1, eVPol);
+    ReadAntennaGain(VgainTopFile, settings1, eVPolTop);
+    ReadAntennaGain(HgainFile, settings1, eHPol);
+    ReadAntennaGain(TxgainFile, settings1, eTx);
 
 }
 //Defining function that reads in TX antenna impedances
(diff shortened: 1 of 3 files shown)


Detector.cc
```diff
@@ -2278,129 +2278,99 @@ inline double Detector::SWRtoTransCoeff(double swr){
...
-    // now below are values that shared in all channels
-    for (int i=0;i<freq_step;i++) { // copy values
-        xfreq[i] = Freq[i];
+inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAntennaType type) {
+
+    // make sure data is being read into the right variables
+    double (*gain)[freq_step_max][ang_step_max];
+    double (*phase)[freq_step_max][ang_step_max];
+    vector<double>* transAnt_databin;
+
```
abigailbishop, line 2288: Can we get a quick comment here summarizing what's happening here?

Detector.cc
```diff
@@ -2278,129 +2278,99 @@ inline double Detector::SWRtoTransCoeff(double swr){
...
     const int N = freq_step;
-    double Transm[N]; 
+    double Transm[N];
     string line;
-    if ( NecOut.is_open() ) {
-        while (NecOut.good() ) {
-            for (int i=0; i<freq_step; i++) {
+    if (! NecOut.is_open() ) 
+      throw runtime_error("Antenna gain file could not be opened: "+filename);
+    while (NecOut.good() ) {
```
abigailbishop, line 2327: Would you be opposed to adding some white space and a comment to each block of code explaining what each section of 5 or so lines are doing? I know these files and I'm struggling to follow how this is reading in the files. Is this iterating over anticipated frequencies and not by lines in the file?

abigailbishop:
Did you get a chance to run AraSim using each of our gain files before and after you made this change to ensure the performance of AraSim doesn't change? This was just changing how the files are read in so the only changes should be in scenarios where the files were being mis-read right?

Detector.cc
marcomuzio: Sure I'll go thru and add some additional comments to the parts I added.

Detector.cc
marcomuzio: Sure, I can do this. I know the way github has laid it out is difficult to follow, but also I only changed the way lines were read-in to fix the problem Justin found. The read-in anticipating the number of frequencies and stuff was already there, which isn't ideal. But I did not attempt to correct this. (In part because I think the number of frequencies is hardcoded, if I recall correctly? So I'm not sure how many things depend on it)

marcomuzio:
> Did you get a chance to run AraSim using each of our gain files before and after you made this change to ensure the performance of AraSim doesn't change? This was just changing how the files are read in so the only changes should be in scenarios where the files were being mis-read right?

I did not check all the gain files. But I added a bunch of random spaces and tabs to the gain file used by the Veff test and made sure that still worked and passed.

Detector.cc
marcomuzio: Added!

Detector.cc
abigailbishop: Ohh okay I didn't realize AraSim was already built to read in by anticipating frequencies. Funky. Thanks!

Detector.cc
abigailbishop: Thank you!

abigailbishop:
Okay that's good to hear! I feel like just a NNU=2500 test would be sufficient to double check that we don't have any mystery problems arising but I'm also just a nervous Nelly so if you're ready to merge this you could probably go for it

PR #122, opened 2024-04-11
Antenna-electronics consistency

marcomuzio:
This update adds new functionality to ensure that the electronics gain model used in a simulation is consistent with the antenna gain model used in that simulation. This is done by calculating the electronics gain on the fly from the corresponding in-situ noise model, antenna gain model, and amplifier noise figure. The calculation itself is outlined in the slides linked below, based on [DocDB 2799](https://aradocs.wipac.wisc.edu/cgi-bin/DocDB/ShowDocument?docid=2799).

This feature is controlled by the `ELECTRONICS_ANTENNA_CONSISTENCY` option. The default setting (`ELECTRONICS_ANTENNA_CONSISTENCY=1`) ensures this consistency whenever an in-situ gain model is used and an in-situ noise model is requested. If this feature is turned off, the in-situ gain amplitude read-in from the corresponding file is left unchanged. The electronics gain phase is left unchanged from what is read-in from the in-situ gain model file. This feature has no impact under other settings configurations. 

A presentation outlining this PR can be found at [DocDB 3063](https://aradocs.wipac.wisc.edu/cgi-bin/DocDB/ShowDocument?docid=3063). Currently, there is a factor of 2 (6 dB in voltage) difference between this calculation and the tabulated values for A1, A2, A4, and A5 (see slides linked). It is not clear where this difference originates from but it is difficult to track down due to the number of different people involved in the calculation of these tables. Moreover, given this calculation agrees with some tables, the error may originate from discrepancies between the noise model data used in this calculation and that used in previous calculations. At least one factor of 2 that was originally in the noise model calculations was later removed by Myoungchul (see changes to `tools/ara_waveform_analyzer.py` in his [`commit 6acf1a8`](https://github.com/MyoungchulK/MF_filters/commit/6acf1a855513e14facdaa8f0496b4e800fa4fd30#diff-aaefc63a6027cd7afc94883fcdcd63a3aa62ae40c2f35f051d85731cd15e7f36)). Note that this change was committed in 22 September 2022, whereas most noise models were generated by or around the noise modeling mini-workshop on 7 July 2022. Corresponding gain models were not generated in earnest until the Fall of 2022 or later. Presumably, this factor was inconsistent between the gain and noise model calculations for all stations except for A3 & PA? But if anyone finds an error in this calculation that would explain the discrepancy, please let me know. That being said merging this PR would provide more transparency to how the electronics gain models are calculated which may still be a net positive.

As part of this update a `GetSplitterFactor` function was added to the `Detector` class. The `ApplySplitterFactor` function was therefore updated to call this new Get function.


Detector.cc
```diff
@@ -1799,6 +1800,18 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
               ReadElectChain(std::string(the_gain_filename), settings1);
               //ReadElectChain("./data/gain/ARA_Electronics_TotalGain_TwoFilters.csv", settings1);
               //ReadElectChain("./data/gain/ARA_Electronics_TotalGainPhase.csv", settings1);
+              if(settings1->ELECTRONICS_ANTENNA_CONSISTENCY==1) {
+                if(settings1->NOISE==1) {
+                  cout << " Recalculating in situ electronic response amplitude to ensure consistency with antenna model used here." << endl;
+                  ReadAmplifierNoiseFigure(settings1); // load amplifier noise figure
+                  CalculateElectChain(settings1); // calculate consistent electronics chain gain amplitude
+                }
+                else {
```
AlanSalcedo, line 1809: So we only calculate it in the fly when you turn on in-situ noise modeling?

marcomuzio: Right, you need the in-situ noise model to do the on-the-fly calculation so I think this is necessary. (Hopefully I got all the right conditions for that)

PR #123, opened 2024-05-08
Correcting Factors of Sqrt(2) in antenna factor calculations.

FlahertyJustinC:
Correcting some factors of sqrt(2) and 1/2 that were missed in Pull Request https://github.com/ara-software/AraSim/pull/110.  The sqrt(2) factors were folded into the electronics gain files for applicable stations, and the factor of 1/2 was a double counting issue in the effective height that has since been resolved.


marcomuzio:
Hey Justin, this looks good to me!

PR #124, opened 2024-05-08
Consolidation of Invert and Apply Antenna and Electronics Functions, and Clean-Up of functional arguments.

FlahertyJustinC:
Consolidated the Invert and Apply Antenna and Electronics factor functions. I accomplished this by adding a boolean in the Apply functions that invert the gain factors and applies a negative sign to the phase factors. The Invert functions still exist, but simply call the Apply functions with the invert boolean enabled. This serves to future-proof the invert functions, as this will always keep them 1-to-1 with the Apply functions. I also cleaned up some of the functional arguments that are no longer used, like the infamous n_trg_slappy and n_trg_pokey.  Additionally, I added descriptions to the Apply and Invert functions for Antenna and Electronics response.


Report.cc
```diff
@@ -4560,108 +4580,78 @@ void Report::ApplyAntFactors_Tdomain (double AntPhase, double heff, Vector &n_tr
...
         if (useInTransmitterMode==true){ 
             phase_current += PI/2;
-            v_amp *= heff * pol_factor;
-            v_amp *= freq/CLIGHT*(Z0/Zr)/4/sqrt(2.);// The factors of two in this line are currently up for debate.  Will be cleaned up in future push - JCF 3/2/2024           
-        }
-        else {
-            // V amplitude
-            v_amp *= heff * pol_factor; 
+            // The factors of two in this line are currently up for debate.  Will be cleaned up in future push - JCF 3/2/2024 
+            v_amp *= pow(freq/CLIGHT*(Z0/Zr)/4/sqrt(2.), amplitudeSign);
```
AlanSalcedo, line 4591: This sqrt(2) might have gone unnoticed. Looks like it was deleted in the gaintoheight() function

Report.cc
```diff
@@ -4560,108 +4580,78 @@ void Report::ApplyAntFactors_Tdomain (double AntPhase, double heff, Vector &n_tr
...
-    vm_bin0 *= heff * pol_factor;
-    vm_bin1 *= heff_lastbin * pol_factor;
+    vm_bin0 *= pow(heff * pol_factor, amplitudeSign);
+    vm_bin1 *= pow(heff_lastbin * pol_factor, amplitudeSign);
     
     if (useInTransmitterMode) {
         // The factors of two in these two lines are currently up for debate.  Will be cleaned up in future push - JCF 3/2/2024  
-        vm_bin0 *= freq/CLIGHT*(Z0/(Zr))/4/sqrt(2.);
-        vm_bin1 *= freq/CLIGHT*(Z0/(Zr))/4/sqrt(2.);
+        vm_bin0 *= pow(freq/CLIGHT*(Z0/(Zr))/4/sqrt(2.), amplitudeSign);
```
AlanSalcedo, line 4627: same here

AlanSalcedo:
I like this consolidation! @marcomuzio are you reviewing this one too?

marcomuzio:
This looks good to me from a quick skim! Thanks for taking a thorough look @AlanSalcedo!

Report.cc
FlahertyJustinC: I left that in for Tx mode as the amplitude is too high without it.  I need to do a second pass on the Tx math when I get the chance.

Report.cc
FlahertyJustinC: Same as my previous comment about the Tx amplitude being too high without this factor.

PR #126, opened 2024-06-28
Restructure Signal+Noise Convolution Functions

abigailbishop:
Cleaned up the part of Report.cc that loads noise waveforms, convolves rays and their signals, and funnels the waveforms through the tunnel diode. I restructured this so I can more easily access the noise-only and signal-only waveforms for noisy-event filtering in a very soon-to-come pull request and to improve comprehensibility. 

The Antenna_r object now has a V_convolved object which contains the signal of all convolved rays without noise. I also fixed the handling of Antenna_r.V_noise so it is now saved properly.

I changed no physics so this PR should not change the scientific outputs of AraSim. The effective volume comparison linked below shows good agreement between the effective volume calculated with simulations from this branch and the `main` branch. 

aeff_wComps_sigNoise_ARA02_wrtmain


Report.cc
```diff
@@ -2024,357 +2027,26 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
...
-                                    }
-                                }
-                            }
+                    // Fill trigger->Full_window and trigger->Full_window_V with noise waveforms
+                    Prepare_Antenna_Noise(debugmode, ch_ID, i, j, k, settings1, trigger, detector);
 
-                            // if we are sharing same noise waveform for first 8 chs and share same noisewaveforms for others, make sure diff chs use diff noise waveforms
-                            else if (settings1->NOISE_CHANNEL_MODE == 2)
-                            {
+                    // currently there is a initial spoiled bins (maxt_diode_bin)
```
marcomuzio, line 2036: What are spoiled bins?

Report.cc
```diff
@@ -2024,357 +2027,26 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
...
-                    // and replace by init : bin + maxt_diode_bin, fin : bin + NFOUR/2
-                    // after we do this for all channels, do trigger check
-
-                    // calculate the bin values for the signal
-                    signal_bin.clear();
-                    signal_dbin.clear();
-                    connect_signals.clear();
+                    // Combine signals from all rays, add noise, and convolve them through the tunnel diode
+                    // cout<<"    "<<i<<" "<<j<<" "<<k<<" :"<<endl;
+                    if (debugmode == 0) Convolve_Signals(
```
marcomuzio: Sorry to nitpick, can you put the command for this if statement on a separate line tabbed in? I think it's more readable that way. Something like:
`if (debugmode == 0)` 
`  Convolve_Signals(&stations[i].strings[j].antennas[k],` 
`       ch_ID, i, settings1, trigger, detector);`

Report.cc
```diff
@@ -4135,250 +3807,515 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
...
+        int this_signalbin = 0;
+        int n_connected_rays = 0;
+        vector <double> V_signal;
+        GetAntennaSignalWF(
+            m, &n_connected_rays, &this_signalbin, BINSIZE, antenna, &V_signal,
+            settings1, trigger, detector);
+
+        // Get the noise-only wf from just this signal-window then convolve 
+        //   through the tunnel diode
+        if (n_connected_rays>0) GetNoiseThenConvolve(
```
marcomuzio: Same nitpick here.

Report.cc
```diff
@@ -4135,250 +3807,515 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
...
+                    BINSIZE, V_signal);
+            }
+        }
 
+    }   // if not the first raysol (all other raysols)
 
 }
 
 
+// Convolve a single signal into the signal array
```
marcomuzio: Does this actually convolve? Is it more like adding in a waveform that will then be convolved with something else later? Maybe theres a way to make the comment or function name more clear?

marcomuzio: (Probably you inherited this problem, but it made me confused for a second)

Report.cc
```diff
@@ -4135,250 +3807,515 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
...
+        noise_ID_index = bin_value / settings1->DATA_BIN_SIZE;
+        noise_wf_index = bin_value % settings1->DATA_BIN_SIZE;
+        V_noise_only->push_back(
+            noise_wf->at( noise_ID[noise_ID_index] ).at( noise_wf_index )
+        );
+    } // end loop over bins
+
+}
+
+// void Report::DoTunnelDiode(
```
marcomuzio: Do we want this function commented out? Is there are reason to keep it?

Report.cc
```diff
@@ -6041,10 +6208,10 @@ double Report::getAverageSNR2(int raysolnum, int station_i, int trig_analysis_mo
         else if (trig_analysis_mode==1) // Noise + signal triggers
             // Estimate average SNR in topmost vpol
             if(stations[station_i].strings[0].antennas[ant_num].V.size()>raysolnum) {
-                int total_bins = stations[station_i].strings[0].antennas[ant_num].V_noise[raysolnum].size();
+                int total_bins = stations[station_i].strings[0].antennas[ant_num].V_noise.size();
```
marcomuzio, line 6198: Sorry I'm having a little trouble following with the diffs on github. What happened to the `V_noise[raysolnum]` part? is there something else looping over the ray solutions?

marcomuzio:
Hey @abigailbishop , thanks for doing this! I did a quick review and things generally look good, just left a few small comments. I don't have any major concerns especially since things seem to be working the same based on your Veff results, but it is definitely a little difficult to follow closely with the gitdiffs (not your fault). Tried my best to check things seems to be put into new functions correctly for a couple of them, but if there's anything specific you think I should review closely let me know! Otherwise I'm happy once the above comments are addressed. :)

abigailbishop:
Wonderful, thanks so much Marco! Besides what you've done already, I'd welcome any comments or suggestions relating to speed or memory consumption if you have any. I think the way I declared variables means they'll be deallocated from memory after each new function ends but I'm not tremendously proficient in c++ that way just yet.

Report.cc
```diff
@@ -4135,250 +3807,515 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
...
+    int BINSIZE, int channel_index, int StationIndex, vector <double> *V_noise_only,
+    Settings *settings1, Trigger *trigger, Detector *detector
+){
+
+    // Clear old noise waveforms
+    V_noise_only->clear();
 
+    // Pick noise waveform
+    vector< vector <double> > *noise_wf;
+    // vector< vector <double> > &noise_wf = *noise_wf_ptr;
```
marcomuzio: Maybe this line can be removed too?

Report.cc
```diff
@@ -4135,250 +3807,515 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
...
+        // If this channel is one of the first 8, use channel specific noise
+        if ( ( GetChNumFromArbChID(detector, channel_index, StationIndex, settings1) - 1 ) < 8) {
+            noise_wf = &trigger->v_noise_timedomain_ch[ 
+                GetChNumFromArbChID(detector, channel_index, StationIndex, settings1) - 1 
+            ];
+        }
+        // This channel is NOT one of the first 8, use same noise for these channels
+        else {
+            noise_wf = &trigger->v_noise_timedomain_ch[
+                8
```
marcomuzio: Can we put this all on one line?

Report.cc
```diff
@@ -4135,250 +3807,515 @@ void Report::ClearUselessfromConnect(Detector *detector, Settings *settings1, Tr
...
+            GetChNumFromArbChID(detector, channel_index, StationIndex, settings1) - 1 
+        ];
+    }
+    // Only use channel-by-channel temperatures to generate noise for the first 8 channels. 
+    // Use the same temperature for the remaining 8.
+    else if ( settings1->NOISE_CHANNEL_MODE==2) {
+        // If this channel is one of the first 8, use channel specific noise
+        if ( ( GetChNumFromArbChID(detector, channel_index, StationIndex, settings1) - 1 ) < 8) {
+            noise_wf = &trigger->v_noise_timedomain_ch[ 
+                GetChNumFromArbChID(detector, channel_index, StationIndex, settings1) - 1
```
marcomuzio: This might be more readable if you define a new index variable and then use that?
`int idx = GetChNumFromArbChID(detector, channel_index, StationIndex, settings1) - 1;`
`noise_wf = &trigger->v_noise_timedomain_ch[idx];`

marcomuzio:
@abigailbishop I went through again looking for memory allocation issues, but didn't find anything that seemed worrying. So I think it's all good on that front!

Report.cc
abigailbishop: No problem! Alan actually told me it's even safer to use curly braces so I added some of those too

Report.cc
abigailbishop: You got it :)

Report.cc
abigailbishop: Yep, done!

Report.cc
abigailbishop: Good idea! D1

Report.cc
abigailbishop: Of course!

Report.cc
abigailbishop: Oh haha I think that is leftover from previous restructuring attempts. I'll get rid of it.

Report.cc
abigailbishop: Yeah these functions were named like this before. I thought about changing the names of the functions but didn't want to make it too confusing for people who've grown to expect these names. I did add some comments to make it more clear, hopefully!

Report.cc
abigailbishop: I changed it from saving the V_noise for each ray to saving the V_noise for the full trace including all rays. This will go hand in hand with the new V_convolved array that includes the summation of just signal from all rays. Previously it would've been tricky to add all rays into one waveform in post-processing because no time array associated with the signal was being saved, so its not simple (as far as I can tell_ to combine all V[ray_sol_cnt] into one big waveform with all rays together

Report.cc
abigailbishop: I'm not sure but that was a comment that was there before. It essentially removes the first ~100 bins (calculated as `settings->MAXT_DIODE / settings->TIMESTEP`) from the noise+signal waveform that was convolved through the tunnel diode presumably because they're a bit funky from being the first samples passed through the tunnel diode?

marcomuzio: Alrighty, no worries. Obviously you can't clarify if it wasn't your comment! I think it's okay as is then.

Report.cc
marcomuzio: That sounds good to me. It sounds like you still account for all the different ray trace solutions, just in a different place. Should be fine!

PR #127, opened 2024-07-06
Added calculation of launch angles of the RF trajectory from the source vertex.

FlahertyJustinC:
Implemented calculation and storage of launch angles of the RF signal from the source vertex to work in conjunction with [AraRoot Pull #68](https://github.com/ara-software/AraRoot/pull/68).  This stores the MCTruth values of the theta and phi launch angles in the Report pointer alongside the MCTruth values of the receiving angles of the signal.  The calculation of the launch angles follows the same framework as the receiving angles, which is a simple conversion of the Cartesian vector into theta and phi.

### DIFF
--- a/Report.cc
+++ b/Report.cc
@@ -4939,7 +4939,8 @@ void Report::GetAngleLaunch(Vector &launch_vector, double &launch_theta, double 
     /*
     2024-07-01 JCF
     Takes the launch vector of the signal and calculates the theta and phi of the launch vector in the 
-    receiveing antenna's coordinate system.
+    station-centric coordinates and returns them in degrees; where theta=0 is along the upward vertical axis,
+    theta=90 is along the horizon, and theta=180 is along the downward vertical axis.
     
     This is necessary because for source reconstruction, we need the launch vector of the RF signal in 
     conjunction with the polarization to find the neutrino trajectory.
--- a/Report.cc
+++ b/Report.cc
@@ -443,7 +443,7 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
     double phi_rec;
     double theta_rec;
 
-    double freq_tmp, heff, antenna_theta, antenna_phi;  // values needed for apply antenna gain factor and prepare fft, trigger
+    double freq_tmp, heff, antenna_theta, antenna_phi, launch_theta, launch_phi;  // values needed for apply antenna gain factor and prepare fft, trigger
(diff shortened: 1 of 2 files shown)


Report.cc
```diff
@@ -4924,6 +4934,22 @@ void Report::GetAngleAnt(Vector &rec_vector, Position &antenna, double &ant_thet
 
 }
 
+void Report::GetAngleLaunch(Vector &launch_vector, double &launch_theta, double &launch_phi) {
+    
+    /*
+    2024-07-01 JCF
+    Takes the launch vector of the signal and calculates the theta and phi of the launch vector in the 
+    receiveing antenna's coordinate system.
```
clark2668: Does it make sense to calculate the _launch_ angle in the frame of the _target_? Wouldn't it make more sense to calculate the launch angle in the frame of the emitter? Or maybe that's what you actually do and the docs are just not quite right?

Report.cc
```diff
@@ -4924,6 +4934,22 @@ void Report::GetAngleAnt(Vector &rec_vector, Position &antenna, double &ant_thet
...
+    
+    This is necessary because for source reconstruction, we need the launch vector of the RF signal in 
+    conjunction with the polarization to find the neutrino trajectory.
+    */
+    
+    launch_theta = launch_vector.Theta() * TMath::RadToDeg();
+    launch_phi = launch_vector.Phi() * TMath::RadToDeg();
+
+}
+
```
clark2668, line 4953: Some documentation about units and convention would be good -- it looks like this is reported in degrees? And with what convention, 0 = up, 180 = down, etc.

clark2668 (changes requested):
Just a couple questions and requests for more clarifying docs.

Report.cc
FlahertyJustinC: Updated the description to have units and conventions for 0 and 180.

Report.cc
FlahertyJustinC: Maybe my description was vague.  This calculates the launch angle theta and phi in station-centric coordinates, while launch angle in the Ray Tracer is calculated relative to the target vector.  I've updated the description of the function to hopefully make this more clear.

FlahertyJustinC:
> Just a couple questions and requests for more clarifying docs.

Hey Brian.  Are my new pushes the changes you're looking for?  If so, could you approve them?  I can't merge until you've approved the changes.

clark2668 (approved):

clark2668:
Whoops, sorry, looks okay to me.

PR #130, opened 2024-07-18
Skip Noisy Events

### LINKED ISSUE #90: Noise only trigger events
The sim saves an event triggered by only noise as a signal event.
We can check the location of the signal center by [SignalBinTime](https://github.com/ara-software/AraSim/blob/c2e1d255a1ae94b0330bcbe6433ca4308febf1f6/Report.cc#L2774C99-L2774C112). And we see the event that signal center is way outside of final WF time range. 
We assume a rare scenario causes this event: the amplitude of noise + signal is too small to be flagged by the trigger. But some parts in the noise pad (`V_total_forconvlv`) happened to have a big amplitude in more than 3 channels.
This event can be removed by the 'proper' solution listed in the attached slide

Related slide:
20231011_sim_bug_corr_ratio (2)

20231011_sim_bug_corr_ratio (3)

abigailbishop:
Occasionally, signal+noise simulations yield (sometimes many) triggered events that don't actually have ray connections to any antennas or don't have sufficient signal felt by any antennas. This leads to noise-event-triggers infiltrating datasets dedicated to neutrino observations and have, as a consequence, inflated our effective volumes and affected our data analysis pipelines. 

This PR resolves this by: 
1. Skipping events that have no ray connections to every antenna in a station.
2. Skipping events that have no antennas observing non-noise signal greater than 0.01 SNR.

I also introduced three new getter functions to the trigger class so when developers need to retrieve the noise mean/rms in the diode/antenna, they don't need to write a new if/else block based on the NOISE_CHANNEL_MODE. These new functions do that for you. 
`GetAntNoise_diodeMean`
`GetAntNoise_diodeRMS`
`GetAntNoise_voltageRMS`

The plot linked below shows the difference between effective volume calculated with simulations from this branch and the `main` branch. 

aeff_wComps_sigNoise_ARA02_wrtmain

Closes #90


Report.cc
```diff
@@ -2411,56 +2267,42 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
...
+                                        trig_bin++;
                                     }
-                                    else if (settings1->NOISE_CHANNEL_MODE == 2)
+                                }
+                            }
+
+                            //else if (settings1->TRIG_ONLY_BH_ON == 0) {
+                            else
+                            {
+                                // other cases, use all possible chs for trigger analysis
```
marcomuzio: What are the other cases here? It's hard to match these if/else statements. Maybe it's worth making it more explicit in this comment.

Report.cc
```diff
@@ -6166,6 +5997,105 @@ vector<double> Report::getHitTimesVector(Detector *detector, int station_i, int
...
+                    // Check if the max (abs) value is greater than 
+                    //   the signals from the other rays
+                    if ( ray_max_signal > max_signal ){
+                        max_signal = ray_max_signal;
+                    }
+                } 
+
+            } // end if (ray solutions exist)
+
+            // If this antenna had a strong signal-only signal,
```
marcomuzio: Is this just that it _had_ a signal really?

Report.cc
```diff
@@ -6166,6 +5997,105 @@ vector<double> Report::getHitTimesVector(Detector *detector, int station_i, int
...
+  
+}
+
+double Report::get_SNR(vector<double> signal_array, vector<double> noise_array){
+    // Returns the SNR of the provided signal compared to the provided noise
+
+    // if signal_array or noise_array have a size of 0, tell stderr
+    if ( signal_array.size() == 0 ){
+        cerr<<"Signal array provided to Report::get_SNR has size 0. ";
+        cerr<<"Function will return inaccurate values."<<endl;
```
marcomuzio, line 6054: Is there a use-case here? Or should this be crashing the code?

Report.cc
```diff
@@ -6166,6 +5997,105 @@ vector<double> Report::getHitTimesVector(Detector *detector, int station_i, int
...
+    }
+    if ( noise_array.size() == 0 ){
+        cerr<<"Noise array provided to Report::get_SNR has size 0. ";
+        cerr<<"Function will return inaccurate values."<<endl;
```
marcomuzio, line 6059: Same here.

Tools.cc
```diff
@@ -73,6 +73,23 @@ double Tools::getMaxMagnitude(vector<double> v) {
 
 }
 
+double Tools::getMaxAbsoluteMagnitude(vector<double> v) {
+    double mag=0.;
+    for (int i=0;i<(int)v.size();i++) {
+        if (v[i]<0){
```
marcomuzio: This seems overly complicated? Why not just use an absolute value?

Trigger.cc
```diff
@@ -168,389 +168,476 @@ void Trigger::SetMeanRmsDiode(Settings *settings1, Detector *detector, Report *r
...
+        meandiode = 0.;
+        rmsdiode = 0.;
+        rmsvoltage = 0.;
+        
+        // make the size of v_noise_timedomain_diode ngeneratedevents (this will be huge!)
+        v_noise_timedomain.resize(ngeneratedevents);  
+        v_noise_timedomain_diode.resize(ngeneratedevents);
+        
+        // Get all `ngeneratedevents` noise waveforms
+        if ( ngeneratedevents > 10 ) {
```
marcomuzio, line 190: Why this condition and an error message? Obviously it's fine, just seems odd?

Trigger.cc
```diff
@@ -168,389 +168,476 @@ void Trigger::SetMeanRmsDiode(Settings *settings1, Detector *detector, Report *r
...
+            Tools::NormalTimeOrdering(settings1->DATA_BIN_SIZE, v_noise);
 
-	for (int m=0; m<settings1->DATA_BIN_SIZE; m++) {
-	    if ( m>=(int)(maxt_diode/TIMESTEP) && m<settings1->DATA_BIN_SIZE ) {
-	    //bwslice_meandiode[j]+=timedomain_output_e[j][m]/((double)ngeneratedevents*((double)NFOUR/2-maxt_diode/TIMESTEP));
-	    meandiode+=v_noise_timedomain_diode[i][m]/((double)ngeneratedevents * ((double)settings1->DATA_BIN_SIZE-maxt_diode/TIMESTEP));
-	    } // end loop over samples where diode function is fully contained 
+            // Convolve signal through the tunnel diode
+            myconvlv(v_noise, settings1->DATA_BIN_SIZE, detector->fdiode_real_databin,v_noise_timedomain_diode[i]);
+            //myconvlv_new(v_noise, settings1->DATA_BIN_SIZE, detector->fdiode_real,v_noise_timedomain_diode[i]);
```
marcomuzio: Probably should remove this line?

Trigger.cc
```diff
@@ -168,389 +168,476 @@ void Trigger::SetMeanRmsDiode(Settings *settings1, Detector *detector, Report *r
...
 
-            v_noise_timedomain[i].push_back(v_noise[m]);    // save pure noise (not diode convlved) waveforms
+            // Save noise waveform and determine mean noise value in this channel
+            for (int m=0; m<settings1->DATA_BIN_SIZE; m++) {
+                
+                // Add contribution from this bin to the mean diode value we're calculating
+                if ( m>=(int)(maxt_diode/TIMESTEP) && m<settings1->DATA_BIN_SIZE ) {
+                    //bwslice_meandiode[j]+=timedomain_output_e[j][m]/((double)ngeneratedevents*((double)NFOUR/2-maxt_diode/TIMESTEP));
```
marcomuzio: Drop this line too?

marcomuzio:
Hey @abigailbishop thanks so much for fixing this! I left a few small comments but I don't have any major concerns, your cross-checks look good to me.

Report.cc
abigailbishop: Fair, I figured out what's going on and added some more comments :) I think the first if block ignores surface channels in the testbed simulation, the second ignore top VPols and top Hpols, and this third case is the normal "check all channels" case.

Report.cc
abigailbishop: Yep, thank you I fixed the comment

Report.cc
abigailbishop: For the signal array, maybe? I never know with AraSim though I haven't seen this error come up while running the ARA02, ARA05, or Phased Array simulations _yet_. I'll add a `throw()` call

Report.cc
abigailbishop: Yeah I can throw an error as well

Tools.cc
abigailbishop: Idk I think I saw it this way somewhere else or I didn't wanna use the cmath::abs() for some reason? I updated it though

Trigger.cc
abigailbishop: Its to cerr because that's how it was before. I think because it's not necessary for the user to know but it was helpful for people monitoring output. 

The condition is because AraSim used to generate hundreds to thousands of noise events and that's why we used to have output that would say something like "Generating noise waveforms... 0%... 10%... --- 100%". These days we only simulate 1 noise waveform, so having the code tell us about the noise waveform generation I think is noisy and unnecessary so I added the condition so that only if a lot of noise waveforms  need to be generated, do we print out the "noise waveform generation progress updates"

Trigger.cc
abigailbishop: Yep I agree. myconvlv_new doesn't even exist. Done.

Trigger.cc
abigailbishop: Yep done :)

Trigger.cc
marcomuzio: Got it, all sounds reasonable to me!

PR #131, opened 2024-08-13
Pa Triggering Update

abigailbishop:
Updated the RMS calculation in the Phased Array Trigger function to use the antenna's simulated RMS and not a default value of 40 millivolts. This has lead to a slight increase in effective volume shown here: 

This was motivated after noticing recent PA simulations yielded a lower effective volume than traditional ARA stations when we expect the PA should have a more sensitive trigger.


marcomuzio:
Hey @abigailbishop, this looks reasonable to me! No comments on my end.

abigailbishop:
Awesome, thanks Marco!

PR #132, opened 2024-08-22
Adding DETECTOR_YEAR parameter to Settings class that gets called when initializing a Detector class.

FlahertyJustinC:
This is needed to specify which year of the SQL database to be used for A1 and A3, as it defaults to the 2011 values without it.


abigailbishop:
Looks good to me, thanks Justin!

marcomuzio:
Thanks @jflaherty13, this looks good to me too! I wonder if we should make the detector year a required argument of that function to avoid this in the future? Also were there any other functions where reloading might cause a problem? 

I guess we need to also make it so that livetime configs have an appropriate year associated with them. Maybe that is a separate PR.

PR #135, opened 2024-09-27
Add action to test ARA3 Veff

abigailbishop:


abigailbishop:
The check is failing because we need AraRoot to run a realistic detector simulation

PR #143, opened 2024-10-11
Flux model sampling bug fix

marcomuzio:
Fixed bug in flux model sampling caused by reading in a flux model file with more than 50 lines (no error thrown previously to catch this). See example below. Made the code a bit more flexible by replacing arrays with vectors. 

Also tried to make code more readable in a few places, removed redundant code, and cleaned up unnecessary block comments. I did not update all functions to take vector arguments rather than arrays, and instead pass the pointer to the vector.

Here is an example of the erroneous distribution of drawn energies (blue) compared to the flux model (orange, in this case `EXPONENT=113` which uses the flux model in `fluxes/equalTriggersPerLnE.dat`).

Screenshot 2024-10-10 at 5 44 18 PM

After this update the distribution of draw energies is as expected:

Screenshot 2024-10-10 at 10 57 31 PM

### DIFF
--- a/Spectra.h
+++ b/Spectra.h
@@ -6,79 +6,62 @@
 #include <string>
 #include "TRandom3.h"
 
+#include<vector>
 
 using namespace std;
 
 class Settings;
-class Spectra {
-
-private:
-  TRandom3 Rand3;
-  double maxflux;   // max flux value
-//  static const int NSPECTRA_MAX=300;  // why need this??
-  static const int E_bin_max = 50;
-  int E_bin;   // initialize # of energy bins (max = 50)
-
-  double dE_bin;
-
-  void GetFlux(string filename);    // read neutrino flux EdNdEdAdt (in GeV) from filename file
-
-//--------------------------------------------------
-//   TGraph *gEdNdEdAdt;   //graph for EdNdEdAdt flux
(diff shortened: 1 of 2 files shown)


Spectra.h
```diff
@@ -6,83 +6,66 @@
...
+    int E_bin;   // initialize # of energy bins (max = 50)
 
-//--------------------------------------------------
-//   TGraph *gEdNdEdAdt;   //graph for EdNdEdAdt flux
-//   TGraph *gE2dNdEdAdt;  //graph for E2dNdEdAdt flux
-// 
-//   TSpline3 *sEdNdEdAdt; //spline of EdNdEdAdt
-//   TSpline3 *sE2dNdEdAdt;    //spline of E2dNdEdAdt
-//-------------------------------------------------- 
+    double dE_bin;
```
abigailbishop: It looks like dE_bin is only used in this function and nowhere else in this script or all of AraSim. Do you think it needs to remain a class variable or can it just live in this function?

abigailbishop:
Thanks for doing this Marco! Looks good to me outside of one minutia comment

Spectra.h
marcomuzio: Ah nice catch! I will change that.

marcomuzio:
Thanks @abigailbishop, implemented that change!

PR #147, opened 2024-10-23
Bug fixes and Primaries cleanup

### LINKED ISSUE #53: Remove hard coded path references to data folder
AraSim has hard coded path references to the `data` folder, for things like antenna patterns, etc. E.g. https://github.com/ara-software/AraSim/blob/master/Detector.cc#L1002

This is bad, and makes it impossible to run the AraSim executable, or anything that depends on the data folder, outside of the `AraSim` directory. We should remove it in favor of an environment variable like `ARASIM_DIR`.

marcomuzio:
I fixed a few bugs in Spectra and in IceModel & EarthModel that were causing issues with reading AraTrees outside the AraSim directory (this also precipitated the need for environmental variables to be defined in the AraSim tests). Also did a big clean up of the Primaries class (removed ~1000 lines from Primaries.cc and fixed formatting).

### DIFF
--- a/.github/workflows/test_birefringence.yml
+++ b/.github/workflows/test_birefringence.yml
@@ -79,5 +79,6 @@ jobs:
           source root/bin/thisroot.sh
           export ARA_UTIL_INSTALL_DIR=`pwd`/AraRoot/
           export ARA_ROOT_DIR=`pwd`/AraRoot/
+          export ARA_SIM_DIR=`pwd`/AraSim/
           export LD_LIBRARY_PATH="/home/runner/work/AraSim/AraSim/libRootFftwWrapper/build/:$ARA_UTIL_INSTALL_DIR/lib:$LD_LIBRARY_PATH"
           bash test/veff_birefringence/run_veff_birefringence_test.sh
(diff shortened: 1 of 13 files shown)


clark2668:
Have you thought about adding something to AraSim.cc, where the first thing it does is check for the environment variable's existence, and fails out if it can't find it? At least check for existence, even better if it can check if it's pointed somewhere sensible...

marcomuzio:
I can do that, does AraRoot do something like that by chance?

clark2668:
No, I don't think so (though it probably should).

marcomuzio:
Alright how's that look?

clark2668:
I think this looks good. I guess you actually tested it and confirmed that when you don't set ARA_SIM_DIR these errors actually get thrown?

marcomuzio:
Ya, that's right. I just set it to an empty string and the error was thrown.

clark2668:
Awesome. Then I say merge away!

clark2668:
Btw, I suspect this would address #53 ?

marcomuzio:
Good call!

PR #148, opened 2024-10-23
Update PA Github Test

abigailbishop:
Updated the energy model for neutrino generation since the EXPONENT=113 was updated. Also made it so the PA test actually runs.


abigailbishop:
This didn't resolve the issue and AraSim is about to change dramatically. Will address this issue in a newer branch and a newer PR

PR #151, opened 2024-10-30
Added data-driven, quality-controlled noise models for A1 full dataset (2014-2023)

Mohammad-Neutrino:


marcomuzio:
Thanks so much for making those changes! I think it's good to merge.

PR #152, opened 2024-10-31
Merging noise models of configs 3 and 4

Mohammad-Neutrino:
A1 detector configs are primarily based on trigger window size. So, the total runs are divided into two groups:

trigger window = 10 (110 ns)
trigger window = 16 (170 ns)

The dataset of 110 ns was separated as
a. 2014 runs: Config1 ( total 451 runs)

The  170 ns dataset was subdivided based on RF blocks and spectral coefficient (sigma) from noise model
b. 2015+2016 runs: Config2 ( total 558 runs)
c. January-July 2018 runs: Config3 ( total 1351 runs)
d. August 2018 and beyond runs: Config4 ( total 14505 runs)

When the average noise models for each config excluding "the runs"  ( total 207 runs) with bad spectral coefficient values (sigma = Nan, chi2/ndf > 3, 10e-6 < sigma < 0) were plotted, it was found that Config3 and Config4 are identical models (apart from some minor differences in a couple of HPol channels). Further check on "the runs" showed that most of them are from Config3. Therefore, Config3 and Config4 are merged and "the runs" are flagged as bad runs in A1 runlogs.


marcomuzio:
Hi @Mohammad-Neutrino why do the noise models for configs 1 & 2 change? And also 3? This doesn't just remove the model for config 4?

Mohammad-Neutrino:
Hey @marcomuzio, noise models looks a bit different especially in the beginning and end because I increased the bandwidth during band passing from 125-875 MHz to 90-900 MHz. The previous filter was decreasing the actual bandwidth (150-850 MHz) by making the gain at 150 MHz very small which should not be the case considering a Butterworth filter. So, they might look a little different but in simulation we will limit the bandwidth anyway and hence, this should not be  problem I guess? Current Config 3 is previous config 4 and the remaining good runs in previous config 3. But again, they look the same except for the start and end.

marcomuzio:
Thanks for the explanation @Mohammad-Neutrino, I think this probably really doesn't matter? The noise models have very little power outside the bandwidth (except for anomalous spikes) anyway. The fallingg edges of the raw model and the one with bandpass applied look very similar. So I'm not sure there's much to be saved. But we can just commit these anyway given I doubt it matters. So please merge if you are happy with these.

PR #155, opened 2024-12-18
On-the-fly gain model bug

marcomuzio:
Updated if statement to load the default gain model when using on-the-fly gain calculation. This is necessary to avoid crashing out if the livetime configuration requested has no corresponding data-driven gain model (which it may not since we have stopped producing them for new livetime configurations).

Also found a bug I apparently inserted in January that didn't actually read-in the default gain models for traditional ARA stations, so fixed that.

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -1786,7 +1786,7 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
           //read the standard ARA electronics
           if(settings1->DETECTOR_STATION > 0){
             char the_gain_filename[500];
-            if(settings1->DETECTOR_STATION_LIVETIME_CONFIG == -1) {
+            if(settings1->DETECTOR_STATION_LIVETIME_CONFIG == -1 || settings1->ELECTRONICS_ANTENNA_CONSISTENCY==1) {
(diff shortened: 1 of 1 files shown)


clark2668:
Uhh, why did we stop producing them for new livetime configs?

marcomuzio:
Because they depend on the antenna model, so it's better to calculate them on-the-fly in simulation.

clark2668:
Right, but I don't understand why we'd want to go to a default. Like, either the user has asked for something sensible, or they haven't. But there shouldn't be a fallback if they've asked for something wrong.

marcomuzio:
I don't think there's an option right now for the user to say "I want to use the data-driven noise models for this livetime config but not its data-driven gain models" except to turn on the on-the-fly calculation. So loading the default is just to avoid a crash when it can't find the gain model. But _some_ gain model needs to be loaded in to get the phase information.

clark2668:
I'm not sure I understand really, but ok.

marcomuzio:
Sorry maybe I'm not really being clear. But just to say, the user hasn't asked for anything "wrong" if they request a particular livetime config (that has an existing noise model) with on-the-fly gain calculation. If they request the data-driven gain model and it doesn't exist this will still crash out.

clark2668:
Ah, okay. Then proceed.

PR #156, opened 2025-01-28
Allow tests to be run from browser

marcomuzio:
Added option to workflow yaml file in order to allow actions to be triggered from the browser following this post: [https://github.com/orgs/community/discussions/58022](https://github.com/orgs/community/discussions/58022)


clark2668:
Makes sense to me!

PR #157, opened 2025-02-03
Fixed merged A5 channels in PA simulation 

marcomuzio:
Fixed merged A5 channels in PA simulation to have correct positions based on correction made in AraRoot. The normal A5 simulation uses the geometry from the AraRoot SQL database and so doesn't need to be addressed directly.

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -1919,20 +1919,20 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
             // Only PA Attached to PA DAQ in DETECTOR_STATION == 1
             if (settings1->DETECTOR_STATION == 2){ // 1 Vanilla VPol from 1 string attached to PA DAQ
                 stations[i].strings.push_back(temp_string);
-                stations[i].strings[1].antennas.push_back(temp_antenna); // A5E 24, A5RF  7, PA  5 (Formally "String 4")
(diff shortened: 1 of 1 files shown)


Detector.cc
```diff
@@ -4097,7 +4098,7 @@ void Detector::ReadAmplifierNoiseFigure(Settings *settings) {
     // define PA channels to be replaced
     vector<int> chans = {5, 10, 11, 12, 13, 14, 15};
     // define corresponding A5 RF channels  
-    vector<int> srcChans = {7, 6, 3, 5, 1, 4, 0};
+    vector<int> srcChans = {7, 6, 3, 4, 0, 5, 1};
```
drparamita, line 4101: @marcomuzio what is being changed here ? what order is this following compared to ` vector<int> `chans = {5, 10, 11, 12, 13, 14, 15};

marcomuzio: `chans` is the PA channel number and `srcChans` is the RF channel number. So this is updating the PA-RF channel mapping. So here RF chan 7 corresponds to PA chan 5, and RF chan 6 corresponds to PA chan 10, for example. 

The PA-electric channel mapping did not change, but the electric-RF channel mapping did which is why this changed.

drparamita: got it ! thanks ! I checked the RF-electric-PA channel map. 
This matches with my findings. 
All changes look good to me. Please merge !

marcomuzio: Thank you!

PR #159, opened 2025-02-09
Updated amplifier noise figure for A1

Mohammad-Neutrino:
Made some changes to get rid of the unphysical values (NF < 1) of A1 noise figure in low frequency region. Also the issue of noise figure being flattened at notch filter region is fixed. Here is a comparison plot

Screenshot 2025-02-08 at 11 05 28 PM


marcomuzio:
Thanks @Mohammad-Neutrino I think this looks good. Did you try running anything with this? My only concern is that the updated file doesn't include 1 GHz, whereas the old one did? Can you add that line? Even if it's the same as the frequency just below it I think that's fine -- obviously it's out of band anyways. But might avoid some complaining from AraSim.

Mohammad-Neutrino:
@marcomuzio Just added 1 GHz data point. I did not run anything with it. Maybe we will get to know when we run simulations. Thanks!

marcomuzio:
Alright thanks! Please merge.

Mohammad-Neutrino:
Merging now. Thanks @marcomuzio !

PR #161, opened 2025-02-10
Remove livetime config checks

### LINKED ISSUE #160: PA livetime config needs expanded range
Throws an error if PA livetime config is greater than 2. There should be 7 livetime configurations available--this is just an outdated check in Settings.

marcomuzio:
Removed the livetime configuration checks in `Settings::CheckCompatibilitiesSettings()` to make code easier to scale in the future. This should be okay because if an invalid livetime configuration is chosen AraSim should crash when it attempts to find the data files corresponding to that livetime configuration. 

Also tried to clean up the code in that area a bit.

### DIFF
--- a/Settings.cc
+++ b/Settings.cc
@@ -887,23 +887,6 @@ int Settings::CheckCompatibilitiesSettings() {
 
     int num_err = 0;
 
-
-    /*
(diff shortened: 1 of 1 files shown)


AlanSalcedo:
Looks good to me

marcomuzio:
Thanks @AlanSalcedo!

PR #162, opened 2025-02-16
Fixed PA amplifier noise figure file.

marcomuzio:
Fixed the PA amplifier noise figure data. Data was taken in the lab at 50 MHz intervals, resulting in a poor interpolation in the previous file version. After talking with Eric, I stole the amplifier NF data from the A5 file in the 400-500 MHz range and then shifted and scaled it to match the PA measurements in that range. This allowed for the peak in the NF in that band to be better modeled.

A comparison of the old PA, new PA, and A5 noise figure data in AraSim is shown below.

Screenshot 2025-02-16 at 1 11 55 AM


drparamita:
@marcomuzio  Now it looks consistent! Plus, that wide curved portion between 400-500 MHz is gone. I think the new PA in this frequency range matches the A5 NF exactly.

marcomuzio:
Okay, I am going to merge this. If Eric tells me this is wrong I will revert.

PR #163, opened 2025-02-18
Improved antenna gain file read-in

marcomuzio:
Updated the way we read in antenna gain files to rely less on hard-coded values, which may or may not correspond to the files in question. The number of frequencies, the initial frequency, the frequency step (`freq_width`), and number of angular steps are now dynamically determined from the first file read in. 

This update requires that all antenna types have a matching set of frequency values. (This was implicitly assumed before but never checked and probably led to some errors...) The exception to this is the transmitter antenna which is hardcoded and so generally unrelated to the other antenna models that are read-in. For this reason some dedicated variables for the transmitter antenna were added. 

Additionally, for safety, the antenna impedance was given its own frequency vector -- but for now we still assume it's the same among all antennas.

Tested and this looks to work for all `ANTENNA_MODE` values (except `ANTENNA_MODE == 2` whose files seem to be missing?). I don't know why the result of the A3 Veff test changed, but maybe those results were due to a bug from the frequency binning mismatches?

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -77,28 +77,15 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
     params.number_of_antennas = 0;
 
     //initialize few params values.
-    params.freq_step = 60;  //The hard-coding of this seems unnecessary, as it limits forces us to have our gain files go beyond 1000 MHz in data, which is out of band for us. - JCF 2/22/2024
-                            //In addition to the hardcoding issue, I think this (potentially) conflicts with the max size of arrays which is set by freq_step_max = 60 in Detector.h - MSM 4/10/2024
-    params.ang_step = 2664;
-    params.freq_width = 16.667;
-    params.freq_init = 83.333;
     params.DeployedStations = 4;
     //end initialize
 
     //Parameters to use if using Arianna_WIPLD_hpol.dat
 
     if (settings1 -> ANTENNA_MODE == 2) {
-        params.freq_step = 238; //60
-        params.ang_step = 2664; //2664;
-        params.freq_width = 5; //16.667;
-        params.freq_init = 15; //83.333;
         params.DeployedStations = 4;
     }
     
-    if (settings1->ANTENNA_MODE == 5 or settings1->ANTENNA_MODE == 6) {
-        params.freq_step = 56; //This is a stop-gap measure to allow the Kansas models ot work in AraSim, which only have data up to 1000 MHz rather than 1066.66 MHz. - JCF 2/22/2024
-    }
-
     //copy freq_width, freq_init in params to Detector freq_width, freq_init
     freq_step = params.freq_step;
     ang_step = params.ang_step;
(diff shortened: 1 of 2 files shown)


Detector.h
```diff
@@ -246,8 +247,11 @@ class Detector {
         //Define impedance and gain for transmitting antenna
         double RealImpedanceTx[freq_step_max];
         double ImagImpedanceTx[freq_step_max];
-        double Txgain[freq_step_max][ang_step_max];
-        double Txphase[freq_step_max][ang_step_max];
+        int Tx_freq_init;
+        int Tx_freq_width;
+        vector<double> TxFreq;
+        vector<vector<double> > Txgain;
```
Mohammad-Neutrino, line 253:
```suggestion
       vector<vector<double> > Txgain;
```

Detector.h
```diff
@@ -227,15 +227,16 @@ class Detector {
...
         double SWRtoTransCoeff(double swr);
         void ReadAntennaGain(string filename, Settings *settings1, EAntennaType type);
-        double Vgain[freq_step_max][ang_step_max];
-        double Vphase[freq_step_max][ang_step_max];
-        double VgainTop[freq_step_max][ang_step_max];
-        double VphaseTop[freq_step_max][ang_step_max];
-        double Hgain[freq_step_max][ang_step_max];
-        double Hphase[freq_step_max][ang_step_max];
-        double Freq[freq_step_max];
+        vector<vector<double> > Vgain;
```
Mohammad-Neutrino, line 230: I am assuming, Vgain and Vphase are for bottom Vpols? It would have been consistent to name them like tops but maybe there are too many (hidden) use throughout AraSim?

Detector.cc
```diff
@@ -2338,109 +2341,153 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
...
+    if(freq_step == -1 || type == eTx) // only reset if it hasn't been read-in yet
+      freq->clear(); 
     
     // check the file opened successfully
     if (! NecOut.is_open() ) 
       throw runtime_error("Antenna gain file could not be opened: "+filename);
-   
+ 
+    // read the first line
+    getline (NecOut, line);
```
Mohammad-Neutrino, line 2358: Is NecOut here and in other places, indicating Nec2 simulated files? Or is it some dummy placeholder name?

Detector.cc
```diff
@@ -7063,42 +7145,40 @@ int Detector::getAntennafromArbAntID( int stationID, int ant_ID){
...
-    // }
-    // //VPol Tx
-    // else if ( ant_m == 0 and useInTransmitterMode) {
-    //     TempRealImpedance = &RealImpedanceTx;
-    //     TempImagImpedance = &ImagImpedanceTx;   
-    // }       
+
+    // this is really dumb but were gonna assume all the impedances have the same frequency
+    // binning and just rewrite this every time a file is read-in. seems okay for now 
+    // and i just don't have the time, someone make this better one day -- MSM 2/18/25
```
Mohammad-Neutrino, line 7151: Love this comment!

Mohammad-Neutrino:
Hi @marcomuzio as always, very detailed work! I didn't test it but if it's working for you, I say we can merge. Thanks for making things better!

Detector.h
marcomuzio: Ya I think that's right. I just kept the names already in AraSim. (Already changed/fixed more than I intended to lol)

Detector.cc
marcomuzio: It's the variable containing the file being read in. I'm not sure why it was given this name originally.

Detector.h
marcomuzio: Why? Isn't this aligned already?

abigailbishop:
I'm not super familiar with this part of AraSim but I don't see anything glaring and it's comforting that all the tests pass. Seems good on my end, thanks Marco!

marcomuzio:
Thanks all!

PR #164, opened 2025-02-19
NFOUR sanity checks 

marcomuzio:
Added two checks on the value of `NFOUR` passed by the user: 1) that `NFOUR` is a power of 2, which comments in the code imply is required (probably for the FFT?) and 2) that `NFOUR` is at least twice the value of `WAVEFORM_LENGTH`. This is enforced since `NFOUR` needs to be large enough to ensure there are not artifacts from the FFT of the waveform and because some vectors have length `WAVEFORM_LENGTH` while others which are read into/out of have length `NFOUR/2`.


Settings.cc
```diff
@@ -887,6 +887,18 @@ int Settings::CheckCompatibilitiesSettings() {
 
     int num_err = 0;
 
+    // ensure that NFOUR is a multiple of 2
+    if (NFOUR & (NFOUR-1) != 0) {
```
abigailbishop, line 891: Did NFOUR need to be a multiple of 2 or a power of 2? I don't remember what but for some reason I think some variable in AraSim needs to be a power of 2

abigailbishop: Oh the error message says a power of 2 as well

Settings.cc
```diff
@@ -115,7 +115,7 @@ outputdir="outputs"; // directory where outputs go
 
   PHASE=90.;            // default : 90 deg phase (it means all imaginary values)
 
-  NFOUR=1024;           // default : 1024, same as in icemc
+  NFOUR=2048;           // default : 2048
```
abigailbishop, line 118: What was the motivation for this update?

abigailbishop:
Left a couple comments but once addressed I think this PR is good to go. Thanks Marco!

Settings.cc
marcomuzio: Oh ya sorry that comment is wrong I will fix that.

Settings.cc
marcomuzio: The default value of `WAVEFORM_LENGTH` is 640 and 640*2=1280 > 1024. So the default values don't current pass these checks. So either we change the default length to 320 (seems pretty short) or we do this

Settings.cc
marcomuzio: Fixed!

marcomuzio:
Thanks @abigailbishop! I think since all these test results have to change anyway, I will wait for the other PR to be merged before doing the final update and merging.

Settings.cc
abigailbishop: Roger that, the update seems good to me!

marcomuzio:
Okay all tests have been updated! @abigailbishop can you take one more quick look when you have a chance to make sure I didn't do anything dumb?

test/veff_birefringence/run_veff_birefringence_test.sh
```diff
@@ -1,10 +1,10 @@
 # run AraSim with the veff setup file
 
-./AraSim test/veff_birefringence/birefringence_setup 0 outputs/. 2>&1 | tee test/veff_birefringence/veff_birefringence_test_output.txt 
+./AraSim test/veff_birefringence/birefringence_setup 2 outputs/. 2>&1 | tee test/veff_birefringence/veff_birefringence_test_output.txt
```
abigailbishop, line 3: Why did you change the run number?

abigailbishop:
I added one question but I think it looks good to go!

test/veff_birefringence/run_veff_birefringence_test.sh
marcomuzio: Ah just becuz the total event weight of triggered events was ~0, so I changed the run to change the seed to get a non-zero total weight rather than increase the number of thrown events.

abigailbishop: Ah makes perfect sense. Thanks!

marcomuzio:
Thanks @abigailbishop!

PR #165, opened 2025-02-21
Set noise event weight to 1

marcomuzio:
Force events to have weight 1 when in noise-only (`TRIG_ANALYSIS_MODE=2`) mode. This also sets `probability` to 1 in case this is ever run when `INTERACTION_MODE!=1`.

### DIFF
--- a/Event.cc
+++ b/Event.cc
@@ -115,6 +115,13 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
         //Report *report_tmp;
 
         Nu_temp = new Interaction (pnu, nuflavor, nu_nubar, n_interactions, icemodel, detector, settings1, primary1, signal, sec1 );
+      
+        // set weights to 1 if this is a noise only simulation
+        if(settings1->TRIG_ANALYSIS_MODE==2) {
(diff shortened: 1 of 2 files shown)


Event.cc
```diff
@@ -115,6 +115,13 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
         //Report *report_tmp;
 
         Nu_temp = new Interaction (pnu, nuflavor, nu_nubar, n_interactions, icemodel, detector, settings1, primary1, signal, sec1 );
+      
+        // set weights to 1 if this is a noise only simulation
```
abigailbishop: Since the settings are passed to the Interaction initialization, can we put this in the interaction initialization to keep all weight and probability calculations in the same place?

abigailbishop:
I left a slightly finicky request but regardless of if you ignore it or fulfill it, this PR looks good to me!

Event.cc
marcomuzio: No you are totally right. I think that was my intention and then I got overwhelmed tracking the event weights haha. Fixed!

marcomuzio:
Thanks @abigailbishop!

PR #166, opened 2025-02-25
First commit towards multitriggers

marcomuzio:
**The multitriggers PR series is a major change to AraSim and will break backwards compatibility. As such it should be treated essentially as a new release.**

This is the first commit towards the merge of the multitriggers PR, which we are attempting to break into a series of a few (slightly) more bite-sized PRs. This PR merges changes to the `Detector`, `Settings`, and `Primaries` classes. This work is due to the tireless efforts of @abigailbishop and @AlanSalcedo.

General information about multitriggers:

This series of PRs is to allow for the reading-in of event lists (eg generated by NuLeptonSim) of primary neutrinos and secondaries which produce additional showers within the detector volume. All waveforms associated with a single primary neutrino are collected and formed into a single, extended trace. The trigger module then scans this trace and determines if there is a trigger. If there is, the data is save and packaged as an output event. After this the trigger module continues to analyze any remaining trace after being shift down the trace by the detector deadtime. This continues until now trace is left to be analyzed.

Multistation coincidences can also be simulated within this framework by using a single global event list for the entire array and determining if the same primary neutrino (including any of its secondaries) causes a trigger in more than one station.

Note that since these event lists are forward modeled, their event weights are 1. This is currently not explicitly enforced, but by construction essentially all events have event weights of ~1. This may be updated later. 

Traditional AraSim simulations can still be run using the usual settings (save for the change that the iterative triggering on a trace is, in principle, still running -- though this is unlikely to cause multiple triggers due to the typically short trace lengths compared to the deadtime). However, old ROOT files produced by AraSim are unlikely to be supported anymore. To read old files, please roll back to a previous commit before this one.

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -45,15 +45,12 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
     NoiseFig_numCh = 16;
 
     for (int i = 0; i < settings1 -> NFOUR / 4; i++) {
-
         freq_forfft.push_back((double) i * freqstep); // even numbers
         freq_forfft.push_back((double) i * freqstep); // odd numbers
-
     }
     for (int i = settings1 -> NFOUR / 4; i < settings1 -> NFOUR / 2; i++) {
         freq_forfft.push_back((double) i * freqstep); // even numbers
         freq_forfft.push_back((double) i * freqstep); // odd numbers
-
     }
     // end of settings freq_forfft
 
@@ -81,9 +78,9 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
     //end initialize
 
     //Parameters to use if using Arianna_WIPLD_hpol.dat
-
-    if (settings1 -> ANTENNA_MODE == 2) 
+    if (settings1 -> ANTENNA_MODE == 2) { 
         params.DeployedStations = 4;
+    }
     
     //copy freq_width, freq_init in params to Detector freq_width, freq_init
     freq_step = params.freq_step;
@@ -121,40 +118,37 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
                             strings.push_back(temp);
                         }
                         cout << "read numner_of_strings" << endl;
-                        //                        Parameters.number_of_strings = atoi( line.substr( line.find_first_of("=") + 1).c_str() );
-                    } else if (label == "antenna_string") {
+                    } 
+                    else if (label == "antenna_string") {
                         string_id++;
                         antenna_id = 0;
                         cout << "read antenna_string" << endl;
-                        //                        if (string_id + 1 >= params.number_of_strings) {
-                        //                            cout<<"Error! Too many strings!"<<endl;
-                        //                        }
-                    } else if (label == "x") {
-                        //strings[string_id].x = atof( line.substr( line.find_first_of("=") + 1).c_str() );
+                    } 
+                    else if (label == "x") {
                         strings[string_id].SetX(atof(line.substr(line.find_first_of("=") + 1).c_str()));
                         cout << "read x : " << (double) strings[string_id].GetX() << " string_id : " << string_id << endl;
-                    } else if (label == "y") {
-                        //strings[string_id].y = atof( line.substr( line.find_first_of("=") + 1).c_str() );
+                    } 
+                    else if (label == "y") {
                         strings[string_id].SetY(atof(line.substr(line.find_first_of("=") + 1).c_str()));
                         cout << "read y : " << (double) strings[string_id].GetY() << " string_id : " << string_id << endl;
-                    } else if (label == "z") {
+                    } 
+                    else if (label == "z") {
                         strings[string_id].antennas.push_back(temp_antenna);
-                        //strings[string_id].antennas[antenna_id].z = atof( line.substr( line.find_first_of("=") + 1, line.find_first_of(",") ).c_str() );
                         strings[string_id].antennas[antenna_id].SetZ(atof(line.substr(line.find_first_of("=") + 1, line.find_first_of(",")).c_str()));
                         strings[string_id].antennas[antenna_id].type = atoi(line.substr(line.find_first_of(",") + 1).c_str());
                         cout << "read z : " << (double) strings[string_id].antennas[antenna_id].GetZ() << " string_id : " << string_id << " antenna_id : " 
                              << antenna_id << " type : " << (int) strings[string_id].antennas[antenna_id].type << endl;
                         antenna_id++;
                         params.number_of_antennas++;
-                        //                        Parameters.number_of_antennas++;
                     }
 
                 }
             }
             testbed.close();
         } 
-        else 
+        else {
             cout << "Unable to open antenna array file !" << endl;
+        }
 
         // testbed version of FlattoEarth_ARA 
         // strings and antennas on the strings use geoid surface!
@@ -170,8 +164,9 @@ Detector::Detector(Settings * settings1, IceModel * icesurface, string setupfile
             theta_tmp = Dist / R1; // assume R1 is constant (which is not)
             phi_tmp = atan2(strings[i].GetY(), strings[i].GetX());
 
-            if (phi_tmp < 0.) 
+            if (phi_tmp < 0.) { 
                 phi_tmp += 2. * PI;
+            }
 
             // set theta, phi for strings.
             strings[i].SetThetaPhi(theta_tmp, phi_tmp);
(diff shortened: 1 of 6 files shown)


clark2668:
So, I'm about to leave a more detailed review, but given this will be a major breaking change, we should freeze a version of this before the change. So there is a "stable" release before the breaks go in.

Detector.cc
```diff
@@ -5994,6 +5994,84 @@ void Detector::getDiodeModel(Settings *settings1) {
...
+    for (int i=0; i < len; i++) {
+        fdiode.push_back( diode_real_fft[i] );
+    }
+
+    delete fdown1;
+    delete fdown2;
+    delete f_up;
+    
+    return fdiode;
+}
```
clark2668, line 5822: Since we're making changes anyway: can you please lint this code to have standard 4-space indentation and proper indentation levels for easy readability? May as well do it right.

Report.cc
```diff
@@ -1298,8 +1298,8 @@ void Report::Connect_Interaction_Detector_V2(Event *event, Detector *detector, R
                                             double V_forfft[stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt]];
                                             double T_forfft[stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt]];
                                             
-                                            cout << stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt] << endl;
-                                            cout << waveform_bin << endl;
+                                            //cout << stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt] << endl;
+                                            //cout << waveform_bin << endl;
```
clark2668: Delete rather than comment out? That or make the cout more descriptive, e.g.
```
printf("Station %d, String %d, Antennas %d, Sol %d : Nnew %d,i,j,k,ray_sol_cnt, stations[i].strings[j].antennas[k].Nnew[ray_sol_cnt]")
```

Settings.cc
```diff
@@ -294,6 +294,8 @@ outputdir="outputs"; // directory where outputs go
   
   WAVEFORM_CENTER = 0; // Default: 0, no offset in waveform centering
 
+  DEADTIME = 0.01*1.E-3; // Default: 0.01 millisecond
```
clark2668: Add docdb reference

Settings.cc
```diff
@@ -818,10 +823,36 @@ void Settings::ReadEvtFile(string evtfile){
...
         evtFile.close();
-	if (NNU == 0 || NNU > EVID.size()){
-	  //	  EVENT_NUM = EVID.size();
-	  NNU = EVID.size();
-	}
+
+        NNU=EVID.size()+1;
+
+        //start counting the number of unique neutrinos
+        if (NNU == 0 || NNU > EVID.size()){
```
clark2668: I'm sorry, can you add some docs about what EVID is storing? And maybe a couple comment lines about what, at a high level, is being done in this loop?

Settings.cc
```diff
@@ -818,10 +823,36 @@ void Settings::ReadEvtFile(string evtfile){
...
+            NNU = 0; //restart from zero
+
+            if (!EVID.empty()) {
+                int interactions_per_nnu_cnt = 1;  // Start the counter at 1 for the first occurrence
+
+                // Iterate through the EVID vector starting from the second element
+                for (size_t i = 1; i < EVID.size(); ++i) {
```
clark2668: Does this have any problems when EVID.size()==1? Or rather, does NNU and interactions_per_nnu_cnt get incremented correctly when this is just one? I don't think I really understand what `interactions_per_nnu_cnt` is doing, so I'm not sure how to assess this. Some more comments (like I suggest above) would be good.

clark2668 (changes requested):
Made some comments and requests for clarification that I think will help. Mostly style and documentation requests.

Settings.cc
AlanSalcedo: @marcomuzio I'll take these. I definitely didn't describe much.

Settings.cc
AlanSalcedo: I'll add more comments and make sure it works. For the moment, interactions_per_nnu_cnt is a counter of all primary + secondary (stochastic losses, decays, ...) associated with the same parent neutrino. All these associated interactions have the same EVID in the input event list.

marcomuzio:
> So, I'm about to leave a more detailed review, but given this will be a major breaking change, we should freeze a version of this before the change. So there is a "stable" release before the breaks go in.

Agreed. Any idea how we should do it? Do you want to just freeze it via a cvmfs build on the latest commit?

Detector.cc
marcomuzio: Sure, I thought about this but figured the diff would be much more difficult to parse. I can work on this.

Report.cc
marcomuzio: Will delete, very happily haha.

clark2668:
> > So, I'm about to leave a more detailed review, but given this will be a major breaking change, we should freeze a version of this before the change. So there is a "stable" release before the breaks go in.
> 
> Agreed. Any idea how we should do it? Do you want to just freeze it via a cvmfs build on the latest commit?

I would "tag" a version on GitHub. And probably we should also freeze a cvmfs version. Some kind of v4.0. So I think both.

Settings.h
```diff
@@ -357,6 +359,7 @@ class Settings
         vector<double> IND_NNU_THETA;
         vector<double> IND_NNU_PHI;
         vector<double> ELAST;
+        vector<int> INT_PER_NNU;
```
abigailbishop: Can we describe this object in a comment on this line like we see for some of the other objects?

clark2668: That is a great idea!

Detector.h
```diff
@@ -467,6 +467,7 @@ class Detector {
         int max_number_of_antennas_station; // maximum number of antennas in a station
 
         void getDiodeModel(Settings *settings1);   // similar with icemc -> anita -> getDiodeModel().  set diode_real and fdiode_real values.
+        vector<double> getDiodeModel(const int len, Settings *settings1);
```
abigailbishop: I know the name is descriptive, but can we add a comment to this line explaining this object like we see for some of the other objects declared in this header?

Detector.cc
marcomuzio: Okay done. The diff is now terrible as promised lolol.

marcomuzio:
FYI @clark2668 we added Primaries.cc to this commit, if you want to take a look.

Settings.cc
marcomuzio: How's that?

clark2668:
Thanks! This is extremely nitpicky, but while I'm here, I generally don't like code like this:
```
                                    if (k == 0 || k == 1) 
                                        stations[i].strings[j].antennas[k].orient = 1;
```

I think it's much clearer as
```
                                    if (k == 0 || k == 1){ 
                                        stations[i].strings[j].antennas[k].orient = 1;
                                    }
```
with brackets. It also makes it more resistant to linting confusion (imo). Is this worth fixing?

Otherwise, I appreciate the linting effort on Detector.cc!

Settings.cc
clark2668: Better, thanks!

Detector.h
marcomuzio: Just updated, lmk if you think this works.

marcomuzio:
@clark2668 donezo!

clark2668:
Thanks @marcomuzio ! I think we're still waiting for some more docs from @AlanSalcedo ?

marcomuzio:
@clark2668 this is ready for final review

Settings.h
abigailbishop: The update looks perfect, thanks!

Detector.h
abigailbishop: Thanks Marco! I'm realizing I don't actually know what "diode model" refers to in physical units. Do you know what parameters of the diode this is modeling and can we add that to the comment?

Detector.cc
```diff
@@ -5888,73 +5667,52 @@ void Detector::getDiodeModel(Settings *settings1) {
         
         diode_real.push_back(0.);   // first puchback 0. value  (this is actually not standard way though works fine)
```
abigailbishop: This wasn't you but there's a typo here. puchback -> pushback

Detector.h
marcomuzio: Uhh that's a good question... my impression is that it's basically the diode impulse response in frequency space? The output of this is what the waveform gets convolved with I think, at least. This function I think just takes that model and evaluates it at the relevant frequency bins.

Detector.cc
abigailbishop: Resolved.

Settings.cc
clark2668: Looks good.

clark2668 (approved):
I'm satisfied. Let's rock and roll!

PR #167, opened 2025-02-27
Second commit towards multitriggers.

marcomuzio:
**The multitriggers PR series is a major change to AraSim and will break backwards compatibility. As such it should be treated essentially as a new release.**

This is the second commit towards the merge of the multitriggers PR, which we are attempting to break into a series of a few (slightly) more bite-sized PRs. The first PR in this series was [PR#166](https://github.com/ara-software/AraSim/pull/166). This PR merges changes to the `Event` and `secondaries` classes. This work is due to the tireless efforts of @abigailbishop and @AlanSalcedo.

General information about multitriggers:

This series of PRs is to allow for the reading-in of event lists (eg generated by NuLeptonSim) of primary neutrinos and secondaries which produce additional showers within the detector volume. All waveforms associated with a single primary neutrino are collected and formed into a single, extended trace. The trigger module then scans this trace and determines if there is a trigger. If there is, the data is save and packaged as an output event. After this the trigger module continues to analyze any remaining trace after being shift down the trace by the detector deadtime. This continues until now trace is left to be analyzed.

Multistation coincidences can also be simulated within this framework by using a single global event list for the entire array and determining if the same primary neutrino (including any of its secondaries) causes a trigger in more than one station.

Note that since these event lists are forward modeled, their event weights are 1. This is currently not explicitly enforced, but by construction essentially all events have event weights of ~1. This may be updated later. 

Traditional AraSim simulations can still be run using the usual settings (save for the change that the iterative triggering on a trace is, in principle, still running -- though this is unlikely to cause multiple triggers due to the typically short trace lengths compared to the deadtime). However, old ROOT files produced by AraSim are unlikely to be supported anymore. To read old files, please roll back to a previous commit before this one.

### DIFF
--- a/Event.cc
+++ b/Event.cc
@@ -120,9 +120,9 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
                 first_vertex_idx = inu_thrown;
             }
             else{
-                // Compute the birth time for secondary interactions (inu_thrown) after its first interaction (first_vertex_idx)
-                // This is determined by dividing the interaction distance by the speed of light (CLIGHT),
-                // giving the travel time from the previous interaction.
+                // Compute the birth time for the current secondary interaction (inu_thrown) after 
+                // its first interaction (first_vertex_idx). This is determined by dividing the interaction 
+                // distance by the speed of light (CLIGHT), giving the travel time from the previous interaction.
                 // Assumes particles travel at ~CLIGHT.
                 double tmp_birth_time = (sec1->get_interaction_distance(first_vertex_idx, inu_thrown, settings1)/CLIGHT);
 
--- a/Event.cc
+++ b/Event.cc
@@ -38,11 +38,11 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
     interaction_cnt_max = n_interactions;
 
     // If the event generation mode is set to 1 (event read-in mode), 
-    // adjust interaction count to the number of interactions per neutrino primary
+    // adjust interaction count to the number of interactions per event
     // calculated in Settings.cc (ReadEvtFile) by looping over events with the same EVID
     if (settings1->EVENT_GENERATION_MODE==1){
 
-        // Retrieve the number of interactions for the current neutrino primary
+        // Retrieve the number of interactions for the current event
         interaction_cnt_max = settings1->INT_PER_NNU[event_num];
 
         // Initialize counter for the cumulative number of interactions
@@ -116,7 +116,7 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
                 // The first interaction is set to have a birth time of 0.0 (reference point)
                 interactions_birth_time.push_back(0.0);
 
-                // Store the index of the first neutrino primary for this event
+                // Store the index of the first interaction for this event
                 first_vertex_idx = inu_thrown;
             }
             else{
--- a/Event.cc
+++ b/Event.cc
@@ -120,7 +120,7 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
                 first_vertex_idx = inu_thrown;
             }
             else{
-                // Compute the birth time for secondary interactions (inu_thrown) of a primary (first_vertex_idx)
+                // Compute the birth time for secondary interactions (inu_thrown) after its first interaction (first_vertex_idx)
                 // This is determined by dividing the interaction distance by the speed of light (CLIGHT),
                 // giving the travel time from the previous interaction.
                 // Assumes particles travel at ~CLIGHT.
--- a/Event.cc
+++ b/Event.cc
@@ -3,8 +3,10 @@
 #include "Settings.h"
 #include "Spectra.h"
 #include "Primaries.h"
+#include "secondaries.hh"
 #include "TMath.h"
 #include "Report.h"
+#include "Constants.h"
 
 ClassImp(Event);
 
@@ -14,6 +16,32 @@ using namespace std;
 Event::Event () {
     //default constructor
     Initialize ();
+}
+
+Event::Event (Event *event){
+
+    Event_type = event->Event_type;
+    inu_thrown = event->inu_thrown;
+    first_vertex_idx = event->first_vertex_idx;
+    interaction_cnt_max = event->interaction_cnt_max;
+    inu_passed = event->inu_passed;
+
+    interactions_birth_time = event->interactions_birth_time;
+    event_ID = event->event_ID;
+
+    pnu = event->pnu;
+
+    nuflavor = event->nuflavor;
+    nuflavorint = event->nuflavorint;
+    n_interactions = event->n_interactions;
+    IsCalpulser = event->IsCalpulser;
+    nu_nubar = event->nu_nubar;
+
+    for (int nu_int_idx=0; nu_int_idx<event->Nu_Interaction.size(); nu_int_idx++){
+        Interaction *save_interaction = new Interaction(event->Nu_Interaction[nu_int_idx]);
+        Nu_Interaction.push_back( *save_interaction );
+    }
+
 }
 
 void Event::Initialize () {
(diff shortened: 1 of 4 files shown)


Event.cc
```diff
@@ -16,6 +18,32 @@ Event::Event () {
     Initialize ();
 }
 
+Event::Event (Event *event){
```
clark2668: What is the purpose of this? Like, can you explain why we need a new object?

Event.cc
```diff
@@ -32,115 +60,144 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
 
     inu_thrown = event_num;
 
-    Choose_Evt_Type (settings1);
+    interaction_cnt_max = n_interactions;
+
+    if (settings1->EVENT_GENERATION_MODE==1){
```
clark2668, line 43: Add comment explaining what is going on. E.g. "If read in mode, then need to loop over and calculate the number of interactions per neutrino primary"

Event.cc
```diff
@@ -32,115 +60,144 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
...
-
-        /*
-        if (settings1->NNU_THIS_THETA==1) {    // set specific theta angle for nnu
-            nnu = primary1->GetThatDirection(settings1->NNU_THETA, settings1->NNU_D_THETA);
-        }
-        else { // nnu angle random
-            nnu = primary1->GetAnyDirection();
-        }
-        */
+            //ASG: Calculate the interactions birth time for secondaries
```
clark2668: Explain what is being done.

secondaries.cc
```diff
@@ -1718,3 +1718,30 @@ double Secondaries::interactionLengthNu(Primaries*primary1,Settings*settings1,do
 	double L=Lcm/100.;//convert cm to m.
 	return L;//meters.
 }
+
+double Secondaries::sec_nnu_interaction_distances(int first_vertex_idx, int inu_thrown, Settings *settings1) {
```
clark2668: Why is this helper needed?

clark2668 (changes requested):
Made some requests for more documentation.

Event.cc
marcomuzio: @abigailbishop can you take this one?

Event.cc
marcomuzio: @AlanSalcedo can you take this one?

Event.cc
marcomuzio: @AlanSalcedo and this one?

secondaries.cc
marcomuzio: @AlanSalcedo probably this one too?

Event.cc
abigailbishop: Yes I will get to it soon!

abigailbishop: This was created because when AraSim saves an event to AraTree2, information from the Event object may be removed depending on the simulation settings. In the multi triggers simulation, we don't want to lose any information (specifically the interaction vector) between saved triggers so I tried adding functionality to save a copy of the event object before information was cleared so I could restore that information after the event is saved to AraTree2.

When I did this, I was under the impression that I needed to write an explicit copying function that carefully handled the interactions vector. @marcomuzio did we determine that my doing this (explicitly defining interaction and event cloning methods) was unnecessary?

Event.cc
abigailbishop: Can you remove your initials from this comment? Your ownership of this line should be captured in the blame and commit histories.

Event.cc
```diff
@@ -32,115 +60,144 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
...
+                interactions_birth_time.push_back(0.0);
+                first_vertex_idx = inu_thrown;
+            }
+            else{
+                double tmp_birth_time = (sec1->sec_nnu_interaction_distances(first_vertex_idx, inu_thrown, settings1)/CLIGHT);
+                interactions_birth_time.push_back(tmp_birth_time);
+            }
+            
+
+            // cout << pnu << endl;
```
abigailbishop: Do we still need these commented out lines? Can they be removed?

abigailbishop: There are more commented out lines below. @AlanSalcedo would you mind going through and trimming out the commented code we definitely won't need anymore?

Event.h
```diff
@@ -29,8 +29,15 @@ class Event {
 
       int inu_thrown; // event number. in case we save triggered events only, this event number could be useful
 
+      int first_vertex_idx; // first vertex index for secondary interactions of a given event
```
abigailbishop: Can we specify in the comment that this "index" corresponds to `inu_thrown`, if I understand this correctly.

abigailbishop: This is nitpicky, but can we clean up the spacing in these comments so they match the spacing used by the other comments?

secondaries.cc
abigailbishop: Super nitpicky, do we need to call this `sec_nnu_interaction_distances` or can it just be `get_interaction_distance`? And can we change `inu_thrown` to `second_index`? It took me a second and reading the code to realize how these parameters were being used.

secondaries.hh
```diff
@@ -230,6 +231,9 @@ string flavors[3]; // the gps path of the anita-lite flight
         double TauEnergyInitial(double ptau_final, double Distance, double z_distance, double density);
 
 	double interactionLengthNu(Primaries*primary1,Settings*settings1,double pnu,int nu_nubar,int currentint);
+
+  double sec_nnu_interaction_distances(int first_vertex_idx, int inu_thrown, Settings *settings1); // calculate distances between secondary interactions
```
abigailbishop: This comment makes it sound like you're calculating the distance between like secondary cascade number 4 and secondary cascade number 5 but, do I understand correctly that this calculates the distance between the primary neutrino's cascade and interaction 4 and then the distance between the primary neutrino's cascade and interaction 5?

Event.cc
marcomuzio: Uhh it depends, I think. I think the compiler will automatically define a copy constructor essentially if all of the member objects have defined copy constructors (and no copy constructor is automatically defined). It's possible we don't need this, but maybe the memory allocation for the Interaction objects makes this necessary.

I guess we should just test it.

marcomuzio: Okay so I just tested this by commenting it out on the full-fledged multitriggers branch (which actually uses this) and the code does not compile without it, so it seems it is not automatically defined by the compiler and is necessary to define here.

Looking [here](https://en.cppreference.com/w/cpp/language/copy_constructor) it looks like the copy constructor is defined as taking an argument of the class-type by reference (i.e. `MyClass::MyClass(const MyClass& myClassObject)`) but according to the error message I'm getting from the compiler we are calling it as `MyClass::MyClass(const MyClass*& myClassObject)`, so a reference to a pointer? I'm guessing probably there was a good reason for this? The compiler does note that a constructor of the form of the copy constructor is declared but the types aren't right. So I think in this case it's just a copy constructor of an unexpected type. I can try playing with it a bit to see if I can get it to work.

marcomuzio: If I recall @abigailbishop you had to print the data inside the event to make sure it was actually being copied, right?

abigailbishop: Thanks so much for this Marco! Yes I did have to print the contents of the event object before copying, after copying, and after clearing along with the contents of the copy of the event before and after the clearing.

marcomuzio: Okay after some testing @abigailbishop and I are fairly confident we can do without this. I have updated the PR to remove this definition.

abigailbishop: Thanks Marco! I'll mark this as resolved then.

secondaries.cc
AlanSalcedo: Mmm... I need perspective here. It's a very general calculation for the distance between interactions from which we later get the "birth" time of the interaction with respect to the primary. 

I don't see something I can use in Tools, Vector, or Interaction classes already. 

We could change the name and leave it more general as suggested by @abigailbishop. The function would still be used to calculate distances between interactions from event lists. 

Secondaries or Interaction classes sound like good places it leave it in. 

Thoughts?

Event.h
AlanSalcedo: You mean between lines? Or in the comments themselves?

AlanSalcedo: Ok I made all comment spacing consistent

Event.cc
clark2668: Thanks!

secondaries.cc
```diff
@@ -1718,3 +1718,30 @@ double Secondaries::interactionLengthNu(Primaries*primary1,Settings*settings1,do
 	double L=Lcm/100.;//convert cm to m.
 	return L;//meters.
 }
+
+double Secondaries::get_interaction_distance(int first_int_idx, int second_int_idx, Settings *settings1) {
```
clark2668, line 1722: This function needs documentation (probably makes more sense to put it in the header file)

secondaries.hh
```diff
@@ -230,6 +231,9 @@ string flavors[3]; // the gps path of the anita-lite flight
         double TauEnergyInitial(double ptau_final, double Distance, double z_distance, double density);
 
 	double interactionLengthNu(Primaries*primary1,Settings*settings1,double pnu,int nu_nubar,int currentint);
+
+  double get_interaction_distance(int first_int_idx, int second_int_idx, Settings *settings1); // Calculate the distance from the primary neutrino’s first interaction to each secondary interaction
```
clark2668: More docs than this plz.

clark2668 (commented):
some more requests

secondaries.cc
clark2668: I mean, just some documentation on what it does is really what I'm asking for. So like, "we need a helper function that calculates the distance between interactions when there are more than one, e.g. when there is a muon traveling and leaving stochastics" -- that would be fine.

clark2668:
I left some additional comments, that I think are probably redundant in part with my earlier ones.

secondaries.cc
clark2668: Okay, looks good.

clark2668 (approved):
LGTM

Event.cc
abigailbishop: This appears resolved

Event.cc
abigailbishop: This also appears resolved

Event.h
abigailbishop: This looks good, thanks!

secondaries.cc
abigailbishop: This appears resolved

secondaries.hh
AlanSalcedo: Changed this function to be more generally "the distance between any two interactions". The assumption that the first vertex is the primary is no longer true. I guess this explanation I could add it where I call the fucntion

Event.h
AlanSalcedo: Mmm. It is inu_thrown when the interaction_cnt == 1:

 //Calculate the interactions birth time for secondaries
            if (interaction_cnt == 1){
                  ...

                // Store the index of the first neutrino primary for this event
                first_vertex_idx = inu_thrown;

secondaries.hh
AlanSalcedo: I added the explanation in the call of the function in Event.cc

Event.cc
```diff
@@ -32,142 +34,145 @@ Event::Event (Settings *settings1, Spectra *spectra1, Primaries *primary1, IceMo
...
-            nuflavorint=1;
-        }
-        else if (nuflavor=="numu"){
-            nuflavorint=2;
-        }
-        else if (nuflavor=="nutau"){
-            nuflavorint=3;
-        }
+            else{
+                // Compute the birth time for secondary interactions (inu_thrown) of a primary (first_vertex_idx)
```
abigailbishop: Since it isn't a given that the `first_vertex_idx` is a primary neutrino / initial cascade can we rephrase this to something like "Compute the birth time for secondary interactions (inu_thrown) after an earlier interaction (first_vertex_idx)"?

Event.h
abigailbishop: Ah then can we alert future developers that this corresponds to an index in the arrays of event parameters stored in the settings class?

abigailbishop: Looks better thanks!

secondaries.hh
abigailbishop: Thanks Alan! Resolving since this is being addressed in another thread.

Event.cc
abigailbishop: Sorry one more, can we say "compute the birth time for the current secondary interaction...". After that it also looks good to me!

PR #169, opened 2025-03-14
Update Phased Array Effective Volume Test from Multitriggers Branch

abigailbishop:
The Phased Array effective test is not presently run on the main branch because the line in the bash script that should run AraSim is commented out, meaning the script that checks that the output is as expected is looking at an old output file that is static and therefore always correct. 

I uncommented this line and found that the test succeeds on the WIPAC Cobalt cluster but fails on the GitHub test. We theorize that there is a memory leak that is overlooked on the Cobalt cluster but catastrophic on GitHub due to different memory architectures. The responsible memory leak has not yet been identified. 

@AlanSalcedo has done a lot of great work to find memory leaks on the multitriggers branch so I would like to check and see if one of Alan's changes has serendipitously resolved this problem. If it hasn't I may immediately delete this PR and branch I've just created, or I will leave it open and update it periodically as more memory leaks may be found on the multitriggers branch before the final merge.


abigailbishop:
This first test failure is understandable given the update to AraSim that enforces NFOUR to be at least double the WAVEFORM_LENGTH was not yet updated for the Phased Array effective volume test. This has been updated in a commit and will be pushed to this PR after I finish checking this version of AraSim's performance on the Cobalt nodes.

abigailbishop:
This doesn't work so I'll close this PR. This simulation runs on the cobalts but continues to fail on GitHub. This time the error is `*** stack smashing detected ***: terminated`

PR #178, opened 2025-04-03
Added consistency check for PA

marcomuzio:
This adds a check to make sure the NFOUR is large enough for the PA, which uses a different internal array size than most stations, which typically use `NFOUR/2`. Compare this for vanilla stations:
https://github.com/ara-software/AraSim/blob/871fe4863d11734b02f7456ec3979f475f1e9255/Report.cc#L2896

to this for the PA:
https://github.com/ara-software/AraSim/blob/871fe4863d11734b02f7456ec3979f475f1e9255/Report.cc#L5100

### DIFF
--- a/Settings.cc
+++ b/Settings.cc
@@ -937,13 +937,15 @@ int Settings::CheckCompatibilitiesSettings() {
     // NFOUR is at least twice this to avoid FFT issues
     const int PA_INTERNAL_LENGTH = 1200/(TIMESTEP*1.e9);  
     if ( DETECTOR==5 && 2*PA_INTERNAL_LENGTH > NFOUR) {
-        cerr<<"NFOUR should be at least twice PA_INTERNAL_LENGTH to avoid FFT artifacts or problems with readout!"<<endl; 
+        cerr<<"NFOUR should be at least twice PA_INTERNAL_LENGTH to avoid FFT artifacts or problems with readout!" 
+            << " Your PA_INTERNAL_LENGTH = " << PA_INTERNAL_LENGTH <<endl; 
         num_err++;
     }
 
     // Also ensure that the requested waveform length isn't longer than the internal length
     if ( DETECTOR==5 && PA_INTERNAL_LENGTH < WAVEFORM_LENGTH) {
-        cerr<<"WAVEFORM_LENGTH cannot be longer than PA_INTERNAL_LENGTH to avoid problems with readout!"<<endl; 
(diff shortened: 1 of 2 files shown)


clark2668:
Does there need to be any formal relationship between BINSIZE and NFOUR?

Settings.cc
```diff
@@ -931,6 +931,24 @@ int Settings::CheckCompatibilitiesSettings() {
         cerr<<"NFOUR should be at least twice WAVEFORM_LENGTH to avoid FFT artifacts or problems with readout!"<<endl; 
         num_err++;
     }
+ 
+    // for PA simulations we need to be careful since BINSIZE isnt the usual NFOUR/2 
+    // but is hardcoded to 1200/(settings1->TIMESTEP*1.e9), so we need to make sure 
+    // NFOUR is at least twice this to avoid FFT issues
+    const int PA_INTERNAL_LENGTH = 1200/(TIMESTEP*1.e9);  
+    if ( DETECTOR==5 && 2*PA_INTERNAL_LENGTH > NFOUR) {
```
clark2668, line 939: shouldn't this be `<NFOUR`?

Settings.cc
```diff
@@ -931,6 +931,24 @@ int Settings::CheckCompatibilitiesSettings() {
...
+        cerr<<"NFOUR should be at least twice PA_INTERNAL_LENGTH to avoid FFT artifacts or problems with readout!" 
+            << " Your PA_INTERNAL_LENGTH = " << PA_INTERNAL_LENGTH <<endl; 
+        num_err++;
+    }
+
+    // Also ensure that the requested waveform length isn't longer than the internal length
+    if ( DETECTOR==5 && PA_INTERNAL_LENGTH < WAVEFORM_LENGTH) {
```
clark2668, line 946: Ditto. Don't you want the conditional reversed?

clark2668 (commented):
Added a couple comments?

Settings.cc
marcomuzio: I don't think so? It's a problem if the internal length is less than half of NFOUR.

Settings.cc
marcomuzio: Same here. Only a problem if the waveform length is longer than the internal length.

Settings.cc
clark2668: Oh, okay, I'm thinking about this the wrong way. Stupid negatives. Then fine. thanks.

Settings.cc
clark2668: Ditto.

clark2668:
Merge away.

PR #179, opened 2025-04-04
store the commit hash

clark2668:
This is a simple PR to store the commit hash as a member of Settings.cc.


marcomuzio:
Looks good to me! That's simpler than I imagined!

abigailbishop:
Same here!

clark2668:
Okay, thanks. I just checked,  it indeed saves to the ROOT file correctly. Merging now.

PR #180, opened 2025-04-09
Extra fixes multitriggers 1

marcomuzio:
Some bug fixes to multitriggers. This is still being worked on.


AlanSalcedo:
The effective volume tests needed to be modified in [here](https://github.com/ara-software/AraSim/pull/180/commits/73f3300eab0af23ff6f01bfdfa0f341fb101f650). We think that a break in previous Veff tests is expected. See how we recover Veff test passing by reverting one line of code in [this commit](https://github.com/ara-software/AraSim/pull/180/commits/389ec7aa53aa56eecedb57798ea294be9fd1847c).

We now make sure that traces have not been modified.

EXAMPLE EVENT **AFTER** THE COMMIT: 
Screenshot 2025-04-09 at 10 16 06 p m


EXAMPLE EVENT **BEFORE** THE COMMIT: 
Screenshot 2025-04-09 at 10 15 43 p m

They look identical, but I have checked twice that they are produced before/after the changes in that commit. 

In summary, the trace wasn't changed significantly or unexpectedly. We still weren't sure why the signals aren't centered in the window. 

Then we produced the same event with a detector configuration (C) number that has no "delay-enable=0" on cable delays for triggering in `AraSim/data/trigger` (e.g. we changed from A3 C2, with cable delays in triggering; to A3 C1, without cable delays in triggering). Here's how the same event looks:

Screenshot 2025-04-09 at 10 12 31 p m


The signal not being centered is due to cable delays added for some detector configurations in triggering, not an issue introduced in multitriggers.

AlanSalcedo:
Regarding the offsets from trigger delays. I went ahead and looked at the `a23_4yr_diffuse` branch of AraSim and produced the same events with and without [cable delays in triggering](https://github.com/ara-software/AraSim/blob/77e97783088652cebc0e89abdbb1cd7230155d94/Report.cc#L2271). 

The shift is present in readout event then! Though the ones in the previous comment seem ~2x more pronounced.

Screenshot 2025-04-09 at 11 16 55 p m
Screenshot 2025-04-09 at 11 17 02 p m

marcomuzio (commented):
This all looks good to me. As far as I'm concerned we can merge this. But maybe @abigailbishop or @clark2668 want to take a look first.

clark2668:
There are too many changes (that seem quite technical) for me to really understand or track. So I don't have any comments here.

Makefile
```diff
@@ -24,6 +24,7 @@ CXX = g++
 
 #Generic and Site Specific Flags
 CXXFLAGS     += $(SYSINCLUDES) $(INC_ARA_UTIL) -DGIT_COMMIT_HASH=\"$(GIT_COMMIT_HASH)\"
+CXXFLAGS		 += -Werror=return-type
```
abigailbishop, line 27: What does this do?

Report.cc
```diff
@@ -612,7 +612,13 @@ void Report::BuildAndTriggerOnWaveforms(
...
     int max_total_bin = (
         (stations[station_index].max_arrival_time - stations[station_index].min_arrival_time) 
-        / settings1->TIMESTEP + settings1->NFOUR *3 + trigger->maxt_diode_bin );    // make more time
+        / settings1->TIMESTEP);
+    
+    // Compute next power of two larger than max_total_bin
+    max_total_bin = (int)std::pow(2, std::ceil(std::log2(max_total_bin)));
+    
+    //make more time
+    max_total_bin += settings1->NFOUR *3 + trigger->maxt_diode_bin;
```
abigailbishop, line 623: Can we get a comment reminding us why we need these extra bins? maxt_diode_bin I can understand (would still be good to have a comment) but I don't remember why we need 3*NFOUR

abigailbishop:
After my two comments are addressed I think this is ready to merge. I agree that this does not appear to change any physics, it just extends arrays including the noise array which results in more RNG calls and that is what changes the Jeff test. Thanks Alan and Marco!

Makefile
AlanSalcedo: @marcomuzio

Report.cc
AlanSalcedo: That "//make more time" comment is what AraSim had already. Let me see if I can understand it.

AlanSalcedo: @abigailbishop what do you think now?

Makefile
AlanSalcedo: I got this: It makes the compiler treat missing or incorrect return statements as errors instead of just warnings.

marcomuzio: Ya exactly. This is how I caught a bug in Primaries which was meant to return double but had no return statement, which is v bad.

marcomuzio: (Also we pretty much have compiler warnings turned off which isn't that great IMO...)

Report.cc
abigailbishop: Is it about having two signals in 2*NFOUR? I just went and looked it up and I almost wonder if its how long we expect a signal to be so we want to add that predicted readout window to the number of bins to investigate

Makefile
abigailbishop: Sounds good, thanks!

Report.cc
AlanSalcedo: Closing this

PR #182, opened 2025-04-10
PA Trigger Update

AlanSalcedo:
Abby's PA trigger update before we diverged with multitriggers from main:

https://github.com/ara-software/AraSim/compare/master...pa_trigger_update


Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+    tmp_noise_RMS.push_back( noise_RMS );
+
+    // Create a 10.7 ns window of waveform to analyze for a trigger-worthy signal
+    //   and move this window along the full waveform, returning the bin
+    //   where a strong enough signal exists, if it does.
+    for (int bin=0; bin < waveform.size()-trigger_window_bins; bin++) {
+
+        // Identify the ray solution that arrived closest to this waveform window's starting bin
+        int Likely_Sol = 0; // Initialize to direct ray 
+        int mindBin = 1.e9; // init big values
```
abigailbishop: Redo this comment to something like `minimum signal bin difference, initialized to an unphysical large number`

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+        int dBin = 0;
```
abigailbishop: Add a comment `signal bin with the minimum signal bin difference`

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+    // The RMS of a vector with a single element is the absolute value of that element.
+    vector <double> tmp_noise_RMS;
+    tmp_noise_RMS.push_back( noise_RMS );
+
+    // Create a 10.7 ns window of waveform to analyze for a trigger-worthy signal
+    //   and move this window along the full waveform, returning the bin
+    //   where a strong enough signal exists, if it does.
+    for (int bin=0; bin < waveform.size()-trigger_window_bins; bin++) {
+
+        // Identify the ray solution that arrived closest to this waveform window's starting bin
```
abigailbishop: I think the ray solution arriving closest to the max value on this string might make more sense? Maybe not for near threshold events but at least for high snr events

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+    //    dsignalBin = abs( trigger_antenna->SignalBin[0] - trigger_antenna->SignalBin[1] );
+    //}
+
+    // If there was a PA trigger, calculate V_mimic and save other trigger info.
+    if ( pa_trigger_bin > -1 ){ 
+
+        // Save the bin the wavefom triggered on.
+        // The first bin of `trigger_waveform` corresponds to the `trig_search_init`
+        //   bin of the full `V_convolved` waveform so we need to adjust the 
+        //   `pa_trigger_bin` accordingly.
```
abigailbishop: Update comment to `by the calculated trigger bin`

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
-
-                // Calculate SNR in this antenna using trace w/o noise since 
-                //   this SNR is used with a signal-only SNR efficiency curve 
-                //   to estimate trigger likelihood
-                avgSnr = get_SNR( 
-                    stations[i].strings[0].antennas[8].V_convolved, 
-                    tmp_noise_RMS);
+                }//end bin
+                my_ch_id ++;
+                //cout<<" Peak Value for ant "<<ant<<" is "<<peakvalue<<endl;
```
abigailbishop: remove this line

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+        CircularBuffer **buffer=new CircularBuffer*[numChan];
+        int trig_window_bin = (int)(settings1->TRIG_WINDOW / settings1->TIMESTEP);  // coincidence window bin for trigger
+        int SCTR_cluster_bit[numChan];
+        for(int trig_j=0;trig_j<numChan;trig_j++) SCTR_cluster_bit[trig_j]=0;
+
+        double *TDR_all_sorted_temp;
+        double *TDR_Vpol_sorted_temp;
+        double *TDR_Hpol_sorted_temp;
+
+        if(settings1->TRIG_SCAN_MODE>1){ // prepare TDR storage arrays and initialize all values to 0
```
abigailbishop, line 5389: We only enter this loop through TRIG_SCAN_MODE 5 so this if/else block is necessary

abigailbishop (commented):
These are mostly comments to myself. 

Tests to run:
Veff: PA, A3, birefringence, pulser
5SA: Read in a multitriggers event list and print out the waveforms. Especially if it includes multiple cascades.

Report.cc
abigailbishop: Done

Report.cc
abigailbishop: Done

Report.cc
abigailbishop: Done

Report.cc
abigailbishop: Done

Report.cc
abigailbishop: Done.

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+
+    // The get_SNR function needs a vector to calculate the noise RMS
+    // The RMS of a vector with a single element is the absolute value of that element.
+    vector <double> tmp_noise_RMS;
+    tmp_noise_RMS.push_back( noise_RMS );
+
+    // Create a 10.7 ns window of waveform to analyze for a trigger-worthy signal
+    //   and move this window along the full waveform, returning the bin
+    //   where a strong enough signal exists, if it does.
+    for (int bin=0; bin < waveform.size()-trigger_window_bins; bin++) {
```
marcomuzio: We need to a check that `trigger_window_bins <= waveform.size()` or this may result in some bad behavior depending on how the type casting works.

marcomuzio: Probably to be safe at the very least we should do:
```
const int waveform_length = (int)waveform.size();
for (int bin=0; bin < waveform_lenght-trigger_window_bins; bin++) {
```

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+                dBin = abs( antenna->SignalBin[m][0] - bin ); //ASG is this what's intended?
+                if ( dBin < mindBin ) {
+                    Likely_Sol = m; 
+                    mindBin = dBin;    
+                }
+            }     
+        } 
+
+        // Scale SNR according to full phased array angular response
+        double arrival_angle_scaling_factor = 2.0 / interpolate(
```
marcomuzio: Maybe this line is clearer if we do:
```
const double SNR0 = 2.0;
const double SNR = interpolate(trigger->angle_PA, trigger->aSNR_PA, // The arrival angle vs SNR curve
            antenna->theta_rec[Likely_Sol][0]*180.0/PI-90.0, // arrival angle to get SNR for //ASG is this what's intended?
            (*(&trigger->angle_PA+1) - trigger->angle_PA) - 1 // len(ang_data) - 1
        );
 const double arrival_angle_scaling_factor = SNR0/SNR;
 ```

File: Report.cc [hunk 9]
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+    //   first 10.7 ns time window with an SNR that triggers the PA.
+    // Return -1 if no 10.7 ns time window triggers the PA
+
+    // Determine number of bins corresponding to a 10.7 ns integration window
+    // 10.7 ns from page 12 of the PA design and instrumentation paper: https://arxiv.org/abs/1809.04573
+    int trigger_window_bins = (10.7E-9) / timestep;
+
+    // The get_SNR function needs a vector to calculate the noise RMS
+    // The RMS of a vector with a single element is the absolute value of that element.
+    vector <double> tmp_noise_RMS;
```
marcomuzio: Avoiding push_backs
`vector <double> tmp_noise_RMS(1, noise_RMS);`

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+
+        // Scale SNR according to full phased array angular response
+        double arrival_angle_scaling_factor = 2.0 / interpolate(
+            trigger->angle_PA, trigger->aSNR_PA, // The arrival angle vs SNR curve
+            antenna->theta_rec[Likely_Sol][0]*180.0/PI-90.0, // arrival angle to get SNR for //ASG is this what's intended?
+            (*(&trigger->angle_PA+1) - trigger->angle_PA) - 1 // len(ang_data) - 1
+        );
+
+        // Get this waveform window and calculate its SNR
+        vector <double> waveform_window;
```
marcomuzio: Avoiding push_backs:
`vector <double> waveform_window(waveform.begin()+bin, waveform.begin()+bin+trigger_window_bins);`

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+        for (int window_bin=bin; window_bin<bin+trigger_window_bins; window_bin++) {
+            waveform_window.push_back( waveform[window_bin] );
+        }
+        double snr = get_SNR(waveform_window, tmp_noise_RMS) * arrival_angle_scaling_factor;
```
marcomuzio: Maybe `eff_snr` is clearer?

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+    // Create a 10.7 ns window of waveform to analyze for a trigger-worthy signal
+    //   and move this window along the full waveform, returning the bin
+    //   where a strong enough signal exists, if it does.
+    for (int bin=0; bin < waveform.size()-trigger_window_bins; bin++) {
+
+        // Identify the ray solution that arrived closest to this waveform window's starting bin
+        int Likely_Sol = 0; // Initialize to direct ray 
+        int mindBin = 1.e9; // init big values
+        int dBin = 0;
+        for (int m=0; m<antenna->ray_sol_cnt; m++) {   // loop over raysol numbers
```
marcomuzio: I wonder if it's more efficient to swap these loops? Or make this one the only loop.

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+
+    // If the antenna we use to trigger the PA doesn't have any ray solutions, 
+    //   do not perform the trigger check.
+    if (trigger_antenna->ray_sol_cnt == 0){
+        return;
+    }
+
+    // Create the signal-only waveform that will be used for triggering
+    vector <double> trigger_waveform;
+    for (int bin=trig_search_init; bin<trigger_antenna->V_convolved.size(); bin++){
```
marcomuzio: Sorry I am being a pain and trying to remove our compiler warnings:
```
const int convolved_len = (int)V_convolved.size();
for (int bin=trig_search_init; bin<convolved_len; bin++){
```

marcomuzio: Wait actually let's skip the loop altogether:
```
vector<double>* V_convolved = &trigger_antenna->V_convolved;
const int convolved_len = (int)V_convolved->size();
vector<double> trigger_waveform(V_convolved->begin()+trig_search_init, V_convolved->begin()+convolved_len);
```

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
-    int raySolNum = 0;
-    bool searchSecondRay = true;
-    bool hasTriggered = false;
+    int dsignalBin = 0;
 
-    // If the antenna we use to trigger the PA doesn't have any ray solutions, 
-    //   do not perform the trigger check.
-    if (stations[i].strings[0].antennas[8].ray_sol_cnt == 0){
-        return;
+    // Determine the bin the PA triggers on. If -1, PA does not trigger.
```
marcomuzio: "If bin=-1, ..."

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+
+    //if (trigger_antenna->ray_sol_cnt == 2){
+    //    dsignalBin = abs( signal_bin[0] - signal_bin[1] ); //original kah
+    //    dsignalBin = abs( trigger_antenna->SignalBin[0] - trigger_antenna->SignalBin[1] );
+    //}
+
+    // If there was a PA trigger, calculate V_mimic and save other trigger info.
+    if ( pa_trigger_bin > -1 ){ 
+
+        // Save the bin the wavefom triggered on.
```
marcomuzio: "waveform"

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+        //   bin of the full `V_convolved` waveform so we need to adjust the 
+        //   `pa_trigger_bin` accordingly.
+        stations[i].Global_Pass = pa_trigger_bin + trig_search_init;
+
+        // Save the V_mimic and corresponding time arrays
+        int my_ch_id = 0;
+        for (size_t str = 0; str < detector->stations[i].strings.size(); str++) {
+            for (size_t ant = 0; ant < detector->stations[i].strings[str].antennas.size(); ant++) {
+                double peakvalue = 0;
+                for (int bin=0; bin<BINSIZE; bin++) {
```
marcomuzio: should run from `0` to `waveformLength`

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
+
+                    int bin_value = stations[i].Global_Pass - BINSIZE/2 + bin;
```
marcomuzio: Should be `int bin_value = stations[i].Global_Pass + waveformCenter - waveformLength/2 + bin;` I believe

Report.cc
```diff
@@ -5082,199 +5082,235 @@ bool Report::isTrigger(double eff){
...
-                avgSnr = get_SNR( 
-                    stations[i].strings[0].antennas[8].V_convolved, 
-                    tmp_noise_RMS);
+                }//end bin
+                my_ch_id ++;
+                //cout<<" Peak Value for ant "<<ant<<" is "<<peakvalue<<endl;
+            }//end ant
+        }//end detector
+
+        // Everything below saves information like in triggerCheckLoop() and saveTriggeredEvent()
```
marcomuzio: I don't really understand all this here and below so I'm not gonna review unless you want me to :)

marcomuzio (changes requested):
Thanks everyone! Apologies some of these are a little annoying

Report.cc
marcomuzio: Hmm I'm not sure? We need to preserve the time ordering?

abigailbishop: This is just for the calculation of signal arrival angle at the antenna and shouldn't be used anywhere else. Are you still worried about time ordering in that case?

abigailbishop: I also checked and the Veff test doesn't change before and after I switch to calculating the arrival angle based on the bin with the max signal

abigailbishop: I've added this in but can still revert it if we don't like it. This ray solution is only being used to calculate the arrival angle of the 10.7ns waveform snippet which is later used to in the signal efficiency calculation. So calculating the arrival angle based on the part of the signal waveform with the most signal seemed more physically representative to me than basing it on the first bin in the 10.7ns window

Report.cc
abigailbishop: I've updated it to this: 
```
    // Create a 10.7 ns window of waveform to analyze for a trigger-worthy signal
    //   and move this window along the full waveform, returning the bin
    //   where a strong enough signal exists, if it does.
    int last_starting_bin = waveform.size()-trigger_window_bins;
    if (waveform.size() <  trigger_window_bin){
        last_starting_bin = waveform.size();
    }
    for (int bin=0; bin < last_starting_bin; bin++) {
```

Report.cc
marcomuzio: Oh sorry I misunderstood. I thought you meant the max value in the full waveform, not just the 10.7 ns window. Okay then I am on board!

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done, but you might hate it all.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done.

Report.cc
marcomuzio: Done. Spot checked if the other changes after this PR was opened were retained but they all seem to be there.

Report.cc
```diff
@@ -5125,382 +5125,435 @@ bool Report::isTrigger(double eff){
...
+    const int convolved_len = (int)V_convolved->size();
+    vector<double> trigger_waveform(V_convolved->begin()+trig_search_init, V_convolved->begin()+convolved_len);
     
     // For phased array, waveform length is 680 ns, but 
     // for PA trigger check only 20 ns around the signal bin.
     // This is to avoid getting the second ray
-    // KAH and ARB are not sure where the 1200 number comes from
+    // KAH and ARB are not sure where the 1200ns window of data was inspired from
     // !!!HEY IF YOU ARE FIXING THIS!!! please update the corresponding check in Settings::CheckCompatibilitiesSettings 
     int BINSIZE = 1200/(settings1->TIMESTEP*1.e9);  // Number of bins (aka datapoints) of data to save
```
marcomuzio, line 5319: Now I'm confused if this is used at all anymore?

Report.cc
abigailbishop: I don't know what this is doing but it looks believable so I'm on board!

abigailbishop:
FYI, as of right now, the ARA3, birefringence, and pulser veffs passed

Machtay (approved):
This looks good to me!

abigailbishop:
All the Veff tests passed except the PA which had to be changed due to a change in RNG. I ran a multitriggers run as well and got this waveform: 
AraOut nu_setup_PA_C2 txt run0 event0

abigailbishop:
I'm feeling good about this and am ready to merge when you are Marco

marcomuzio:
> All the Veff tests passed except the PA which had to be changed due to a change in RNG. I ran a multitriggers run as well and got this waveform: AraOut nu_setup_PA_C2 txt run0 event0

I like how this clearly didn't just trigger on the brightest pulse!! Looks great

marcomuzio (approved):
Merge away!

PR #183, opened 2025-04-15
Guarantee that max_total_bin > V_signal.size() after padding

AlanSalcedo:
`max_total_bin` should be equal (or larger) than the size of the full waveform. That is `V_signal.size()` from the funtion on Report.cc `Combine_Waveforms()`.

We were doing `max_total_bin = max_arrival_bin - min_arrival_bin`, which is the time between arrivals in bins. It doesn't reach the end of the waveform. We needed to add the size of a signal which is `BINSIZE = NFOUR/2`. 

In other parts if the code, when there's two ray solutions per interaction, people added `2*BINSIZE = NFOUR`. We mimic that addition on `max_total_bin` for safety.

`V_signal.size()` then get's padded to the next power of two at the end of `Combine_Waveforms()` for further FFTs. You'll see `max_total_bin` padded to that accordingly:

```
    int max_total_bin = (
        (stations[station_index].max_arrival_time - stations[station_index].min_arrival_time) 
        / settings1->TIMESTEP);
        / settings1->TIMESTEP) + settings1->NFOUR;

    // Compute next power of two larger than max_total_bin
    max_total_bin = (int)std::pow(2, std::ceil(std::log2(max_total_bin)));
```


clark2668:
I'm not sure I have the expertise to evaluate this....

Report.cc
```diff
@@ -612,14 +612,15 @@ void Report::BuildAndTriggerOnWaveforms(
     // to save time, use only necessary number of bins
     int max_total_bin = (
         (stations[station_index].max_arrival_time - stations[station_index].min_arrival_time) 
-        / settings1->TIMESTEP);
+        / settings1->TIMESTEP) + settings1->NFOUR;
```
abigailbishop, line 618: Can we get a comment here or above this integer assignment explaining why you've added NFOUR

abigailbishop:
This sounds good, thanks for doing this! Have you researched all references to `max_total_bin` to make sure increasing this number doesn't cause out of bounds indexing later in the code? Like checking that all for loops over `max_total_bin` will respond well with this update?

abigailbishop:
Since this will change the results of the V_eff test, are you planning to update those tests with this PR as well?

AlanSalcedo:
> This sounds good, thanks for doing this! Have you researched all references to `max_total_bin` to make sure increasing this number doesn't cause out of bounds indexing later in the code? Like checking that all for loops over `max_total_bin` will respond well with this update?

I just checked the instanced of `max_total_bin`. No loops look concerning. There's only one where we do the trigger check up from the signal bin up to 'max_total_bin - trig_window_bin' which makes sense and I don't expect it to be a problem.

AlanSalcedo:
@abigailbishop I added the comment you asked for and changes to Veff tests.

Report.cc
abigailbishop: Looks good thanks!

abigailbishop:
Sound good. Thanks Alan- merge away!

PR #186, opened 2025-04-17
Add V_convolved and V_noise to Antenna_r::clear

abigailbishop:
These are intermediate waveform objects not needed in light data_save_modes


AlanSalcedo:
@abigailbishop looks good Abby!

Report.cc
```diff
@@ -267,7 +267,16 @@ void Antenna_r::clear_useless(Settings *settings1) {
   // If the user requests that less information be stored in the output file, 
   //   clear some information here from the Antenna_r object
 
-  if (settings1->DATA_SAVE_MODE == 1) {
+  // DATA_SAVE_MODE meanings from Settings.h
+  // 0 : save all information which are generated during the processes. 
+  //     * Remove nothing *
+  // 1 : light mode, remove most of data before the final global trigger
```
marcomuzio: I think the difference between mode 1 and 2 here is sort of unclear to me? Ik you probably didn't write this

marcomuzio (commented):
I think this should work! To be sure can you run 1 veff test just to be safe?

Report.cc
abigailbishop: How's it now?

marcomuzio: Thanks thats clearer!

PR #190, opened 2025-04-24
Add event weights to AraSim event list Production and ReadIn

abigailbishop:
All Veff tests passed (A3, PA, pulser, and birefringence. I also generated an event list and read this event list in two separate runs of AraSim to ensure that the written event matched the read-in event. I also tested the reading of multiple multi triggers event lists all of which worked. 

There was also a stray `inu_thrown++` at the end of the Event class initializer that I removed. It may have been there so that the first event thrown had an `inu_thrown` of 1 which could be more intuitive than 0, however since we access event list settings parameters with `inu_thrown`, we need this value to not be shifted by 1 for accurate data retrieval.


AlanSalcedo:
Looks good to me

abigailbishop:
Thanks!

PR #192, opened 2025-05-02
Fix bug in getting signal to apply PA trigger check to

marcomuzio:
This fixes a bug in `Report::Get_Brightest_Interaction` that was due to `*brightest_event[0]` actually getting the 0th integer and then dereferencing it, rather than the intended dereferencing of the pointer to the array and _then_ getting its 0th integer element. To correct this array dereferencing was used with `brightest_event[0][0]`.

This affects the PA Veff test and requires us to update it.


marcomuzio:
####  Testing summary
1. A3 veff test: passed ✅ 
3. pa veff test: passed ✅ (updated in this PR)
4. pulser veff test: passed ✅

abigailbishop:
If you tested it and it works well, then this looks good to me!

PR #193, opened 2025-05-06
Added check for ARA_SIM_DIR variable

marcomuzio:
This PR adds a check for the existence and sanity of the `ARA_SIM_DIR` environmental variable. When this isn't defined or pointing to a sensible location it can cause crashes (with unhelpful error messages).

This also required a change in how a global variable is defined in `IceModel.h` which would cause a crash upon including since the conversion of the `nullptr` (which `getenv` returns when a variable is undefined) into a `string` would cause a crash before the check could be run in the main code.

### DIFF
--- a/AraSim.cc
+++ b/AraSim.cc
@@ -46,7 +46,21 @@ void save_useful_event(int, UsefulIcrrStationEvent*, UsefulAtriStationEvent*, do
 string outputdir="outputs";
 
 int main(int argc, char **argv) {   // read setup.txt file
-    
+
+    // check that the environmental variable needed to find data files
+    // is defined before we do anything
+    if(!getenv("ARA_SIM_DIR")) {
+        throw runtime_error("The environmental variable ARA_SIM_DIR must be set!");
+    }
+    // check if a data file can be loaded to ensure this variable
+    // points to a reasonable place
+    {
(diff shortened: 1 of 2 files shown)


marcomuzio:
####  Testing summary
1. A3 veff test: passed ✅ 
3. pa veff test: passed ✅ 
4. pulser veff test: passed ✅

AraSim.cc
```diff
@@ -46,7 +46,21 @@ void save_useful_event(int, UsefulIcrrStationEvent*, UsefulAtriStationEvent*, do
 string outputdir="outputs";
 
 int main(int argc, char **argv) {   // read setup.txt file
-    
+
+    // check that the environmental variable needed to find data files
+    // is defined before we do anything
+    if(!getenv("ARA_SIM_DIR")) {
```
clark2668, line 52: Does this `!` work, or do you need to explicitly check for the nullptr?

clark2668 (commented):
One small comment

AraSim.cc
marcomuzio: Ya it does, I checked. I think it converts to false anyway.

clark2668: Okay. Then gtg.

clark2668:
Gtg then.

PR #194, opened 2025-05-11
Change EVID to 64 bit unsigned int

marcomuzio:
Updated the data type for EVIDs in event lists to a 64 bit unsigned int, since the old `int` was not large enough to store some event IDs.

PR #195, opened 2025-05-11
Increment Settings ClassDef

marcomuzio:
Forgot to do this when changing the EVID data in [PR#194](https://github.com/ara-software/AraSim/pull/194).


clark2668:
gtg

PR #196, opened 2025-05-14
Defining event_ID for EVENT_GENERATION_MODE not 1

AlanSalcedo:
When not reading the event lists, event ID should be the number of the event being simulated.


abigailbishop:
Looks good to me!

marcomuzio (approved):
Makes sense to me!

PR #199, opened 2025-06-04
Reassign noise models to match new A23 configurations

abigailbishop:
We also removed old gain files that are not needed anymore as gain is now calculated on the fly. 

For posterity, here are the commands I ran and some slides that show the mapping of old to new noise models

```
mkdir old
mv sigmavsfreq_A_{2..3}* old
cp old/sigmavsfreq_A_2_config_2.csv sigmavsfreq_A_2_config_1.csv
cp old/sigmavsfreq_A_2_config_3.csv sigmavsfreq_A_2_config_2.csv 
cp old/sigmavsfreq_A_2_config_4.csv sigmavsfreq_A_2_config_3.csv 
cp old/sigmavsfreq_A_2_config_4.csv sigmavsfreq_A_2_config_4.csv 
cp old/sigmavsfreq_A_2_config_6.csv sigmavsfreq_A_2_config_5.csv 
cp old/sigmavsfreq_A_2_config_7.csv sigmavsfreq_A_2_config_6.csv 
cp old/sigmavsfreq_A_2_config_7.csv sigmavsfreq_A_2_config_7.csv 
cp old/sigmavsfreq_A_3_config_1.csv sigmavsfreq_A_3_config_1.csv
cp old/sigmavsfreq_A_3_config_5.csv sigmavsfreq_A_3_config_2.csv
cp old/sigmavsfreq_A_3_config_5.csv sigmavsfreq_A_3_config_3.csv
cp old/sigmavsfreq_A_3_config_3.csv sigmavsfreq_A_3_config_4.csv
cp old/sigmavsfreq_A_3_config_4.csv sigmavsfreq_A_3_config_5.csv
cp old/sigmavsfreq_A_3_config_6.csv sigmavsfreq_A_3_config_6.csv
cp old/sigmavsfreq_A_3_config_8.csv sigmavsfreq_A_3_config_7.csv
cp old/sigmavsfreq_A_3_config_8.csv sigmavsfreq_A_3_config_8.csv
cp old/sigmavsfreq_A_3_config_9.csv sigmavsfreq_A_3_config_9.csv
```

Capture d’écran, le 2025-06-04 à 11 54 03
Capture d’écran, le 2025-06-04 à 11 54 07

PR #202, opened 2025-06-16
Change sinc interpolator extrapolation to 0

marcomuzio:
This PR changes the sinc interpolator used in `Report::PropagateSignal` to extrapolate to signals to 0, rather than to the first or final value in the array. The previous behavior can introduce "steps" into the signal trace, especially when many signal traces are summed (like when secondaries and different ray trace solutions are combined).

The A3 Veff test was also updated due to changes in the A3 config 2 data-drive noise model (really just config-redefinition) in PR #199.

**Here is an example of the behavior before the fix for a 1e20.5 eV tau neutrino ~700 m from PA.**

With saturation (to 435 mV):

Screenshot 2025-06-16 at 1 21 11 AM

Without saturation:

Screenshot 2025-06-16 at 1 28 13 AM
    

**After the fix, the signal looks like the following (NB the trigger bin likely has changed).**

With saturation (to 435 mV):

Screenshot 2025-06-16 at 1 23 58 AM

Without saturation:

Screenshot 2025-06-16 at 1 26 09 AM

### DIFF
--- a/Settings.cc
+++ b/Settings.cc
@@ -101,7 +101,7 @@ outputdir="outputs"; // directory where outputs go
 
   WHICHPARAMETERIZATION=0;  //
 
-  SIMULATION_MODE=1;    // default freq domain simulation
+  SIMULATION_MODE=1;    // default time domain simulation
 
   USE_PARAM_RE_TTERM_TABLE=1; // default: use the interpolation table to get Param_RE_TTerm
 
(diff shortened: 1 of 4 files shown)


marcomuzio:
####  Testing summary
1. A3 veff test: passed ✅  (updated in this PR)
3. pa veff test: passed ✅ 
4. pulser veff test: passed ✅

clark2668:
I think this look ok

PR #203, opened 2025-07-01
Beam pattern read-in 

jacobweiler:
This pull request allows for beam patterns to be read-in from defined paths in the setup file for VPOL, VTOP, and HPOL in ANTENNA_MODE=6. 

Example usage:
In a setup file add in the following lines:
ANTENNA_MODE=6 // (DEFAULT=0) Antenna to be used
VPOL_BEAMPATTERN="path/to/beam/pattern" // (DEFAULT=CustomFileinAra)" Filepath to read in any gain pattern from setup file for VPOL
VTOP_BEAMPATTERN="path/to/beam/pattern" // (DEFAULT=CustomFileinAra)
HPOL_BEAMPATTERN="path/to/beam/pattern" // (DEFAULT=CustomFileinAra) Filepath to read in any gain pattern from setup file for HPOL

### DIFF
--- a/RaySolver.cc
+++ b/RaySolver.cc
@@ -1122,6 +1122,16 @@ void RaySolver::Solve_Ray (Position &source, Position &target, IceModel *antarct
       ns = 1.326;
       nd = 1.78;
       nc = 0.0202;
+    } else if (settings1->RAY_TRACE_ICE_MODEL_PARAMS == 41){ 
+      // UNL Modified (PA model) with upward systematics (see ice model 40). Related slide: https://aradocs.wipac.wisc.edu/0035/003525/002/5SA_Ice_Model_Uncertainties.pdf 
+      ns = 1.40968; // 1.326 + 0.08368
+      nd = 1.788;   // 1.78 + 0.00800
+      nc = 0.0206;  // 0.0202 + 0.00040  
+    } else if (settings1->RAY_TRACE_ICE_MODEL_PARAMS == 42){ 
+      // UNL Modified (PA model) with upward systematics (see ice model 40). Related slide: https://aradocs.wipac.wisc.edu/0035/003525/002/5SA_Ice_Model_Uncertainties.pdf 
+      ns = 1.166;   // 1.326 - 0.16000
+      nd = 1.774;    // 1.78 - 0.00600
+      nc = 0.0155;  // 0.0202 - 0.00470  
     } else if (settings1->RAY_TRACE_ICE_MODEL_PARAMS == 50){ // AC model for birefringence: https://arxiv.org/abs/2110.09015
       ns = 1.35;
       nd = 1.78;
--- a/Settings.h
+++ b/Settings.h
@@ -293,7 +293,9 @@ class Settings
         //11 : Moore's Bay Model 2
         //20 : Byrd (Ebimuna (1983))
         //30 : Mizuho (Ebimuna (1983))
-              //40: UNL Modified (PA model). Related slide: https://aradocs.wipac.wisc.edu/docs/0022/002222/001/inIceMC_Hughes_A5locations_10222020.pdf
+        //40: UNL Modified (PA model). Related slide: https://aradocs.wipac.wisc.edu/docs/0022/002222/001/inIceMC_Hughes_A5locations_10222020.pdf
+        //41: UNL Modified (PA model) with upward systematics. Related slide: https://aradocs.wipac.wisc.edu/0035/003525/002/5SA_Ice_Model_Uncertainties.pdf
+        //42: UNL Modified (PA model) with downward systematics. Related slide: https://aradocs.wipac.wisc.edu/0035/003525/002/5SA_Ice_Model_Uncertainties.pdf
         
         int WAVEFORM_LENGTH; // the number of samples in the waveform length for V_mimic and UsefulAtriStationEvent, default: 64/2*20 = 640
 


Detector.cc
```diff
@@ -2300,10 +2300,10 @@ inline void Detector::ReadAllAntennaGains(Settings *settings1){
         VgainTopFile = string(getenv("ARA_SIM_DIR"))+"/data/antennas/realizedGain/ARA_TVpol_RealizedGainAndPhase_Copol_Kansas2024.txt";
         HgainFile = string(getenv("ARA_SIM_DIR"))+"/data/antennas/realizedGain/ARA_Hpol_RealizedGainAndPhase_Copol_Kansas2024.txt";         
     }
-    else if (settings1->ANTENNA_MODE == 6) { //Adding antenna mode for custom gains.
-        VgainFile = string(getenv("ARA_SIM_DIR"))+"/data/antennas/realizedGain/ARA_BVpol_RealizedGainAndPhase_Copol_Custom.txt";
-        VgainTopFile = string(getenv("ARA_SIM_DIR"))+"/data/antennas/realizedGain/ARA_TVpol_RealizedGainAndPhase_Copol_Custom.txt";
-        HgainFile = string(getenv("ARA_SIM_DIR"))+"/data/antennas/realizedGain/ARA_Hpol_RealizedGainAndPhase_Copol_Custom.txt";         
+    else if (settings1->ANTENNA_MODE == 6) { //Adding antenna mode for custom gains (takes full path).
```
AlanSalcedo: Could you change BEAMPATTERN to GAIN_FILE or something like that? The gain file contains the phase as well as the beam pattern.

AraSim.cc
```diff
@@ -73,6 +73,9 @@ int main(int argc, char **argv) {   // read setup.txt file
     cout<<"MOOREBAY : "<<settings1->MOOREBAY<<endl;
     cout<<"EXPONENT : "<<settings1->EXPONENT<<endl;
     cout<<"DETECTOR : "<<settings1->DETECTOR<<endl;
+    cout<<"VPOL_BEAMPATTERN : "<<string( settings1->VPOL_BEAMPATTERN )<<endl;
```
AlanSalcedo: Could you drop the printing of these paths in AraSim.cc? Or have them printed only when ANTENNA_MODE = 6. Otherwise, it will always print the string you initialized, right? Even if I use other ANTENNA_MODE other than 0.

Settings.cc
```diff
@@ -727,6 +730,36 @@ void Settings::ReadFile(string setupfile) {
               else if (label == "ANTENNA_MODE"){
                   ANTENNA_MODE = atoi(line.substr(line.find_first_of("=") + 1).c_str());
               }
+              else if (label == "VPOL_BEAMPATTERN"){
+                size_t pos = line.find("="); // Variable position
```
AlanSalcedo: Since you're repeating similar code for each gain file, could you make it a function and call it? e.g.

`else if (label == "VPOL_BEAMPATTERN") {
    VPOL_BEAMPATTERN = extract_setup_string(line);
}
`

And have the `extract_setup_string` function defined somewhere below in Settings.cc?

Settings.cc
```diff
@@ -46,6 +46,9 @@ outputdir="outputs"; // directory where outputs go
 
  SIGMA_SELECT=0; // when in SIGMAPARAM=1 case, 0 : (default) use mean value, 1 : use upper bound, 2 : use lower bound
 
+HPOL_BEAMPATTERN="./data/antennas/realizedGain/ARA_dipoletest1_output.txt"; // Default to original Ara Data
```
AlanSalcedo: Could you align these lines with the rest?

AlanSalcedo (changes requested):

AlanSalcedo:
@jacobweiler nice work Jacob! I left a few comments and requested reviews from others.

marcomuzio:
Sorry just catching up here. So the idea is that we are replacing the custom antenna gain file with a "beam pattern" file?  But it's basically just the antenna gain still? (Sorry explain to me like I am 3)

AraSim.cc
```diff
@@ -73,6 +73,9 @@ int main(int argc, char **argv) {   // read setup.txt file
...
+    cout<<"VTOP_BEAMPATTERN : "<<string( settings1->VTOP_BEAMPATTERN )<<endl;
```
marcomuzio: I don't think you need the type casts here? Isn't `settings1->VTOP_BEAMPATTERN` already a string? Same for all the lines around here.

AraSim.cc
```diff
@@ -114,6 +117,9 @@ int main(int argc, char **argv) {   // read setup.txt file
     cout<<"DETECTOR : "<<settings1->DETECTOR<<endl;
     cout<<"POSNU_RADIUS : "<<settings1->POSNU_RADIUS<<endl;
     cout << "EVENT_GENERATION_MODE: " << settings1->EVENT_GENERATION_MODE << endl;
+    cout<<"VPOL_BEAMPATTERN : "<< string( settings1->VPOL_BEAMPATTERN )<<endl;
```
marcomuzio: Same comment.

Settings.cc
```diff
@@ -727,6 +730,36 @@ void Settings::ReadFile(string setupfile) {
               else if (label == "ANTENNA_MODE"){
                   ANTENNA_MODE = atoi(line.substr(line.find_first_of("=") + 1).c_str());
               }
+              else if (label == "VPOL_BEAMPATTERN"){
+                size_t pos = line.find("="); // Variable position
+                size_t comment_pos = line.find("//", pos); // Comment Position
+                string fileloc = (comment_pos != string::npos) ? line.substr(pos + 1, comment_pos - pos - 1) : line.substr(pos + 1);
```
marcomuzio: Can you explain the logic here? What is the form of the line you expect in the setup file?

Settings.cc
```diff
@@ -727,6 +730,36 @@ void Settings::ReadFile(string setupfile) {
               else if (label == "ANTENNA_MODE"){
                   ANTENNA_MODE = atoi(line.substr(line.find_first_of("=") + 1).c_str());
               }
+              else if (label == "VPOL_BEAMPATTERN"){
```
marcomuzio: Do we want some sort of warning if these are set but the antenna mode isn't 6?

marcomuzio (changes requested):
Looks pretty good to me! Just a few small questions and suggestions.

AlanSalcedo:
> Sorry just catching up here. So the idea is that we are replacing the custom antenna gain file with a "beam pattern" file? But it's basically just the antenna gain still? (Sorry explain to me like I am 3)

Yeah, I think the idea is to have the custom antenna model be referenced via a gain file path (same file format and type) in the setup file.

marcomuzio:
> > Sorry just catching up here. So the idea is that we are replacing the custom antenna gain file with a "beam pattern" file? But it's basically just the antenna gain still? (Sorry explain to me like I am 3)
> 
> Yeah, I think the idea is to have the custom antenna model be referenced via a gain file path (same file format and type) in the setup file.

Ah okay that makes sense. I guess I'm wondering a bit about the use case here. Is this for a genetis thing where a lot of different (and short lived) beam patterns are going to be tested?

AlanSalcedo:
> > > Sorry just catching up here. So the idea is that we are replacing the custom antenna gain file with a "beam pattern" file? But it's basically just the antenna gain still? (Sorry explain to me like I am 3)
> > 
> > 
> > Yeah, I think the idea is to have the custom antenna model be referenced via a gain file path (same file format and type) in the setup file.
> 
> Ah okay that makes sense. I guess I'm wondering a bit about the use case here. Is this for a genetis thing where a lot of different (and short lived) beam patterns are going to be tested?

Yes, I think so. @jacobweiler is that what you had in mind?

jacobweiler:
> > > > Sorry just catching up here. So the idea is that we are replacing the custom antenna gain file with a "beam pattern" file? But it's basically just the antenna gain still? (Sorry explain to me like I am 3)
> > > 
> > > 
> > > Yeah, I think the idea is to have the custom antenna model be referenced via a gain file path (same file format and type) in the setup file.
> > 
> > 
> > Ah okay that makes sense. I guess I'm wondering a bit about the use case here. Is this for a genetis thing where a lot of different (and short lived) beam patterns are going to be tested?
> 
> Yes, I think so. @jacobweiler is that what you had in mind?

Yes this is for GENETIS to allow for scripts to input in the filepaths of converted uan files in the loop. There was a way before in AraSim to input in different gain file, but it would be a little messy/dangerous as I'd be writing over a single file quite a bit. 

I will work through all these changes soontm. Thanks for the comments.

Settings.cc
jacobweiler: An example way to input in the file path for an antenna, like VPOL, is to have a line in the settings like:
`VPOL_GAIN_FILE="/users/PAS1977/jacobweiler/GENETIS/test_repos/HPOL/Run_Outputs/2025_02_11_steph_ferrite_test/txt_files/a_1.txt"// (DEFAULT=CustomFileinAra)" Filepath to read in any gain pattern from setup file for VPOL`

I'm not sure if I should add this to an example setup file or not.

marcomuzio: I think what worries me is that this method of reading in the file path depends critically on the setup file having exactly the right syntax. So we should add some sort of error check/warning for the user or find a more robust method. I could see someone easily deleting the comment from that line.

jacobweiler: I agree, I changed it so it is just looking for the path between quotes after variable definition and added error throwing if the format is wrong. I think it should be better now.

Settings.cc
```diff
@@ -46,6 +46,9 @@ outputdir="outputs"; // directory where outputs go
 
  SIGMA_SELECT=0; // when in SIGMAPARAM=1 case, 0 : (default) use mean value, 1 : use upper bound, 2 : use lower bound
 
+ HPOL_GAIN_FILE="./data/antennas/realizedGain/ARA_dipoletest1_output.txt"; // Default to original Ara Data
+ VTOP_GAIN_FILE="./data/antennas/realizedGain/ARA_bicone6in_output.txt"; // Default to original Ara Data
+ VPOL_GAIN_FILE="./data/antennas/realizedGain/ARA_bicone6in_output.txt"; // Default to original Ara Data
```
Machtay: So just to make sure, if someone just wants to change one of these, they don't have to pass the path to the others even in this setting right? If so then I think things are good to go.

jacobweiler: It defaults to what gain files are in ANTENNA_MODE == 0 in Detector.cc, though it seems like the path has slightly changed since I wrote this a year ago... so I will go and try to get that in line with current the current method soontm

If it doesn't see any defined path in the setup file, it should just default to those gains. Again, if anyone thinks other gain files are more apt for this let me know and I will change it. I just put the gain files that I'd want default when evolving antennas in GENETIS, but I could have also picked the wrong mode based on older GENETIS setup scripts (ie, maybe antenna_mode=1 is better?)

marcomuzio: Actually, this reminds me that these default file paths should use an environmental variable to avoid issues running outside of the AraSim directory. Here's an example:

https://github.com/ara-software/AraSim/blob/2593a75dd7be450be0ae439d5580069590e5b98d/Detector.cc#L871

Settings.cc
```diff
@@ -727,6 +730,24 @@ void Settings::ReadFile(string setupfile) {
               else if (label == "ANTENNA_MODE"){
                   ANTENNA_MODE = atoi(line.substr(line.find_first_of("=") + 1).c_str());
               }
+              else if (label == "VPOL_GAIN_FILE") {
+                  VPOL_GAIN_FILE = ParseFilePath(line);
+                  if (VPOL_GAIN_FILE.empty()) {
+                      std::cerr << "Warning: could not parse path from line: " << line << std::endl;
```
marcomuzio: Thanks for updating this! I think this should help a lot. I wonder if we want to just crash the code in this case and throw an error? Warnings that aren't load tend to be ignored. Also it would probably be helpful for the user to add to the message the expected format.

Settings.cc
marcomuzio: Thanks! Made some comments in a new thread.

Settings.cc
jacobweiler: Okay I've added this and added an example of what the input should be.

Settings.cc
jacobweiler: This should also be fixed now.

Settings.cc
```diff
@@ -727,6 +730,30 @@ void Settings::ReadFile(string setupfile) {
...
+                      std::cerr << "Example Input: VPOL_GAIN_FILE=\"path/to/gain/file.txt\"" << std::endl;
+                      std::abort();
+                  }
+              }
+              else if (label == "VTOP_GAIN_FILE") {
+                  VTOP_GAIN_FILE = ParseFilePath(line);
+                  if (VTOP_GAIN_FILE.empty()) {
+                      std::cerr << "Warning: could not parse path from line: " << line << std::endl;
+                      std::cerr << "Example Input: VTOP_GAIN_FILE=\"path/to/gain/file.txt\"" << std::endl;
+                      std::abort();
```
marcomuzio: Did you test this? Do you know what `std::abort()` looks like for the user?

jacobweiler: Yes. If inputted wrong it will pop up with something that looks like this:
```
Warning: could not parse path from line: VPOL_GAIN_FILE=/users/PAS1977/jacobweiler/GENETIS/test_repos/HPOL/Run_Outputs/2025_02_11_steph_ferrite_test/txt_files/a_1.txt// (DEFAULT=CustomFileinAra)" Filepath to read in any gain pattern from setup file for VPOL
Example Input: VPOL_GAIN_FILE="path/to/gain/file.txt"
Aborted
```
If this isn't how you guys normally do this, I can change it to the typical way. 

After retesting I got annoyed with the Antenna Mode error check going off when it's not supposed to, so I have fixed that as well.

marcomuzio: Thanks, that seems fine! Usually I throw a runtime error but this seems to work as well.

marcomuzio (approved):
Thanks for making those changes! I think this all looks good on my end.

Settings.cc
Machtay: Sweet, yeah that's what I was wondering about when I saw that. Looks good to merge on my end.

Machtay (approved):

PR #205, opened 2025-07-10
Ensure PA always triggers in noise only mode

marcomuzio:
This PR adds additional logic to ensure triggers are forced when in noise-only mode for PA. It does not affect how the noise-only case is handled for the normal ARA trigger. 

The trigger is forced by hardcoding the SNR to a large value where the trigger efficiency is 1. Other changes were made to ensure thrown events are not skipped due to a lack of ray trace solutions (from the neutrinos that are generated and then ignored). The current PA trigger implementation is not capable of simulating real RF triggers from noise (since it is based on the signal-only SNR).

### DIFF
--- a/AraSim.cc
+++ b/AraSim.cc
@@ -725,7 +725,7 @@ void save_event_data(
     for (int i=0; i< report->stations.size(); i++) {
         
         // check the total global trigger passed
-        if (report->stations[i].Global_Pass) {
+        if (report->stations[i].Global_Pass >= 0) {
 
             event->inu_passed = Events_Passed;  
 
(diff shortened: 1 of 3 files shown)


marcomuzio:
####  Testing summary
1. A3 veff test: passed ✅ 
3. pa veff test: passed ✅ 
4. pulser veff test: passed ✅

abigailbishop (commented):
Looks good! I haven't done the due diligence of checking that you caught all the Global_Pass in all scripts but if the tests all pass then hopefully you did. 

Do you have an example of a noise waveform? I don't remember the full logic of waveform development using Global_Pass but I just wanna make sure that if there's some pre_trigger that we don't pull some garbage from noise[ global_pass - pre_trigger_window] = noise[ 0 - some_number ] = noise[ some negative number ] = something_out_of_memory

marcomuzio:
Something seems off about the waveforms here, but I'm not sure it's worth figuring out right now. Going to close this for now, we can refer to this PR if we want to finish this later.

PR #207, opened 2025-09-26
Exception for noise only sims

AlanSalcedo:


marcomuzio (approved):
Once you've confirmed this passes the Veff tests, I think its good to go.

PR #209, opened 2025-10-09
Addition of Uzair's analytic ray tracer

AlanSalcedo:
This PR adds the analytic ray tracer developed by Uzair L. An old branch is located here: https://github.com/ara-software/AraSim/compare/master...AnalyticRaytracing

More information about how it works in his thesis: https://kuscholarworks.ku.edu/entities/publication/6d616d78-1a72-41f3-bfcc-735048fb611f

All effective volume tests pass locally.

### DIFF
--- a/IceRayTracing.cc
+++ b/IceRayTracing.cc
@@ -1,0 +1,3039 @@
+/*
+  This is the IceRayTracing namespace. Author: Uzair Latif 
+  released under GPL3.
+*/
+
+#include "IceRayTracing.hh"
+
+void IceRayTracing::SetA(double &A){
+  IceRayTracing::A_ice=A;
+}
+
+void IceRayTracing::SetB(double &B){
+  IceRayTracing::B_ice=B;
+}
+
+void IceRayTracing::SetC(double &C){
+  IceRayTracing::C_ice=C;
+}
+
+/* Get the value of the B parameter for the refractive index model */
+double IceRayTracing::GetB(double z){
+  z=fabs(z);
+  double B=0;
+
+  B=IceRayTracing::B_ice;
+
+  // if(z<=IceRayTracing::TransitionBoundary){
+  //   B=-0.5019;
+  // }else{
+  //   B=-0.448023;
+  // }
+
+  return B;
+}
+
+/* Get the value of the C parameter for the refractive index model */
+double IceRayTracing::GetC(double z){
+  z=fabs(z);
+  double C=0;
+
+  C=IceRayTracing::C_ice;
+
+  // if(z<=IceRayTracing::TransitionBoundary){
+  //   C=0.03247;
+  // }else{
+  //   C=0.02469;
+  // }
+
+  return C;
+}
+
+
+/* Get the value of refractive index model for a given depth  */
+double IceRayTracing::Getnz(double z){
+  z=fabs(z);
+  return IceRayTracing::A_ice+IceRayTracing::GetB(z)*exp(-IceRayTracing::GetC(z)*z);
+}
+
+/* E-feild Power Fresnel coefficient for S-polarised wave which is perpendicular to the plane of propogation/incidence. This function gives you back the reflectance. The transmittance is T=1-R */
+double IceRayTracing::Refl_S(double thetai){
+
+  double Nair=1;
+  double Nice=IceRayTracing::Getnz(0); 
+  double n1=Nice;
+  double n2=Nair;
+
+  double sqterm=sqrt(1-pow((n1/n2)*(sin(thetai)),2));
+  double num=n1*cos(thetai)-n2*sqterm;
+  double den=n1*cos(thetai)+n2*sqterm;
+  double RS=(num*num)/(den*den);
+
+  if(std::isnan(RS)){
+    RS=1;
+  }
+  return (RS);
+}
+
+/* E-feild Power Fresnel coefficient for P-polarised wave which is parallel to the plane of propogation/incidence. This function gives you back the reflectance. The transmittance is T=1-R */
+double IceRayTracing::Refl_P(double thetai){
+
+  double Nair=1;
+  double Nice=IceRayTracing::Getnz(0); 
+  double n1=Nice;
+  double n2=Nair;
+
+  double sqterm=sqrt(1-pow((n1/n2)*(sin(thetai)),2));
+  double num=n1*sqterm-n2*cos(thetai);
+  double den=n1*sqterm+n2*cos(thetai);
+  double RP=(num*num)/(den*den);
+  if(std::isnan(RP)){
+    RP=1;
+  }
+  return (RP);
+}
+
+/* The temperature and attenuation model has been taken from AraSim which also took it from here http://icecube.wisc.edu/~araproject/radio/ . This is basically Matt Newcomb's icecube directory which has alot of information, plots and codes about South Pole Ice activities. Please read it if you find it interesting. */
+
+/* Temperature model:The model takes in value of depth z in m and returns the value of temperature in Celsius.*/
+double IceRayTracing::GetIceTemperature(double z){
+  double depth=fabs(z);
+  double t = 1.83415e-09*pow(depth,3) + (-1.59061e-08*pow(depth,2)) + 0.00267687*depth + (-51.0696 );
+  return t;
+}
+
+/* Ice Attenuation Length model: Takes in value of frequency in Ghz and depth z and returns you the value of attenuation length in m */
+double IceRayTracing::GetIceAttenuationLength(double z, double frequency){
+
+  double t =IceRayTracing::GetIceTemperature(z);
+  const double f0=0.0001, f2=3.16;
+  const double w0=log(f0), w1=0.0, w2=log(f2), w=log(frequency);
+  const double b0=-6.74890+t*(0.026709-t*0.000884);
+  const double b1=-6.22121-t*(0.070927+t*0.001773);
+  const double b2=-4.09468-t*(0.002213+t*0.000332);
+  double a,bb;
+  if(frequency<1.){
+    a=(b1*w0-b0*w1)/(w0-w1);
+    bb=(b1-b0)/(w1-w0);
+  }
+  else{
+    a=(b2*w1-b1*w2)/(w1-w2);
+    bb=(b2-b1)/(w2-w1);
+  }
+  double Lval=1./exp(a+bb*w);
+  return Lval;
+}
+
+/* Setup the integrand to calculate the attenuation */
+double IceRayTracing::AttenuationIntegrand (double x, void * params) {
+
+  double *p=(double*)params;
+
+  double A0=p[0];
+  double Frequency=p[1];
+  double L=p[2];
+
+  double Integrand=(A0/IceRayTracing::GetIceAttenuationLength(x,Frequency))*sqrt(1+pow(tan(asin(L/IceRayTracing::Getnz(x))) ,2));
+  return Integrand;
+}
+
+/* Integrate over the integrand to calculate the attenuation */
+double IceRayTracing::IntegrateOverLAttn (double A0, double Frequency, double z0, double z1, double Lvalue) {
+  gsl_integration_workspace * w= gsl_integration_workspace_alloc (1000);
+
+  double result, error;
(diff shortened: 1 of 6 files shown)


IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
+/*
+  This is the IceRayTracing namespace. Author: Uzair Latif 
+  released under GPL3.
+*/
+
+#include "IceRayTracing.hh"
+
+void IceRayTracing::SetA(double &A){
```
marcomuzio: Why does this need to be passed by reference? This seems weird for a setter function. I guess all the setters have this issue.

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+void IceRayTracing::SetB(double &B){
+  IceRayTracing::B_ice=B;
+}
+
+void IceRayTracing::SetC(double &C){
+  IceRayTracing::C_ice=C;
+}
+
+/* Get the value of the B parameter for the refractive index model */
+double IceRayTracing::GetB(double z){
```
marcomuzio: Seems like this could be cleaned up? Doesn't seem to need the argument since that functionality isn't currently supported.

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  //   B=-0.5019;
+  // }else{
+  //   B=-0.448023;
+  // }
+
+  return B;
+}
+
+/* Get the value of the C parameter for the refractive index model */
+double IceRayTracing::GetC(double z){
```
marcomuzio: Same here

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+}
+
+void IceRayTracing::SetB(double &B){
+  IceRayTracing::B_ice=B;
+}
+
+void IceRayTracing::SetC(double &C){
+  IceRayTracing::C_ice=C;
+}
+
```
marcomuzio: Why is there no`GetA` function?

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  //   C=0.02469;
+  // }
+
+  return C;
+}
+
+
+/* Get the value of refractive index model for a given depth  */
+double IceRayTracing::Getnz(double z){
+  z=fabs(z);
```
marcomuzio: I don't love the absolute value here. We should have a clear convention for z. If negative z values aren't allowed then an error should be raised.

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  if(std::isnan(RP)){
+    RP=1;
+  }
+  return (RP);
+}
+
+/* The temperature and attenuation model has been taken from AraSim which also took it from here http://icecube.wisc.edu/~araproject/radio/ . This is basically Matt Newcomb's icecube directory which has alot of information, plots and codes about South Pole Ice activities. Please read it if you find it interesting. */
+
+/* Temperature model:The model takes in value of depth z in m and returns the value of temperature in Celsius.*/
+double IceRayTracing::GetIceTemperature(double z){
```
marcomuzio: Is this a reimplementation of the one inside `IceModel`? Can we call that instead?

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  double depth=fabs(z);
+  double t = 1.83415e-09*pow(depth,3) + (-1.59061e-08*pow(depth,2)) + 0.00267687*depth + (-51.0696 );
+  return t;
+}
+
+/* Ice Attenuation Length model: Takes in value of frequency in Ghz and depth z and returns you the value of attenuation length in m */
+double IceRayTracing::GetIceAttenuationLength(double z, double frequency){
+
+  double t =IceRayTracing::GetIceTemperature(z);
+  const double f0=0.0001, f2=3.16;
```
marcomuzio: Is this a reimplementation of the one inside `IceModel`? Can we call that instead?

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  gsl_integration_qags (&F, zlimit[0], zlimit[1], 0, 1e-7, 1000,
+                        w, &result, &error);
+
+  // printf ("result          = % .18f\n", result);
+  // printf ("estimated error = % .18f\n", error);
+  // printf ("intervals       = %zu\n", w->size);
+
+  gsl_integration_workspace_free (w);  
+
+  return fabs(result);
```
marcomuzio: Why the absolute value? Is it just cuz the attenuation is positive and there's some issue with the direction of integration?

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+}
+
+/* Calculate the total attenuation for each type of ray */
+double IceRayTracing::GetTotalAttenuationDirect (double A0, double frequency, double z0, double z1, double Lvalue) {
+  z0=fabs(z0);
```
marcomuzio: Same complaint about the depth convention.

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  z1=fabs(z1);
+  return IceRayTracing::IntegrateOverLAttn(A0,frequency,z0,z1,Lvalue);
+}
+
+double IceRayTracing::GetTotalAttenuationReflected (double A0, double frequency, double z0, double z1, double Lvalue) {
+  z0=fabs(z0);
```
marcomuzio: Here too

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  z1=fabs(z1);
+  return IceRayTracing::IntegrateOverLAttn(A0,frequency,z0,0.000001,Lvalue) + IceRayTracing::IntegrateOverLAttn(A0,frequency,z1,0.000001,Lvalue);
+}
+
+double IceRayTracing::GetTotalAttenuationRefracted (double A0, double frequency, double z0, double z1, double zmax, double Lvalue) {
+  z0=fabs(z0);
```
marcomuzio: here too

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+double IceRayTracing::ftimeD(double x,void *params){
+
+  struct IceRayTracing::ftimeD_params *p= (struct IceRayTracing::ftimeD_params *) params;
+  double A = p->a;
+  double B = p->b;
+  double C = p->c;
+  double Speedc = p->speedc;
+  double L = p->l;
+
+  return (1.0/(Speedc*C*sqrt(pow(IceRayTracing::Getnz(x),2)-L*L)))*(pow(IceRayTracing::Getnz(x),2)-L*L+(C*x-log(A*IceRayTracing::Getnz(x)-L*L+sqrt(A*A-L*L)*sqrt(pow(IceRayTracing::Getnz(x),2)-L*L)))*(A*A*sqrt(pow(IceRayTracing::Getnz(x),2)-L*L))/sqrt(A*A-L*L) +A*sqrt(pow(IceRayTracing::Getnz(x),2)-L*L)*log(IceRayTracing::Getnz(x)+sqrt(pow(IceRayTracing::Getnz(x),2)-L*L)) );
```
marcomuzio: this has got to be broken up. this is ridic

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  struct IceRayTracing::ftimeD_params *p= (struct IceRayTracing::ftimeD_params *) params;
+  double A = p->a;
+  double B = p->b;
+  double C = p->c;
+  double Speedc = p->speedc;
+  double L = p->l;
+
+  //integral sec(sin^(-1)(L/(A + B e^(C x)))) dx = (log((A + B e^(C x)) (sqrt((A^2 + 2 A B e^(C x) + B^2 e^(2 C x) - L^2)/(A + B e^(C x))^2) + 1)) - (A log(A sqrt(A^2 - L^2) sqrt((A^2 + 2 A B e^(C x) + B^2 e^(2 C x) - L^2)/(A + B e^(C x))^2) + B sqrt(A^2 - L^2) e^(C x) sqrt((A^2 + 2 A B e^(C x) + B^2 e^(2 C x) - L^2)/(A + B e^(C x))^2) + A^2 + A B e^(C x) - L^2))/sqrt(A^2 - L^2) + (A C x)/sqrt(A^2 - L^2))/C;
+
+  return (log((A + B*exp(C*x))*(sqrt((A*A + 2*A*B*exp(C*x) + B*B*exp(2*C*x) - L*L)/((A + B*exp(C*x))*(A + B*exp(C*x))) ) + 1)) - (A*log(A*sqrt(A*A - L*L)*sqrt((A*A + 2*A*B*exp(C*x) + B*B* exp(2*C*x) - L*L)/(( A + B*exp(C*x))*(A + B*exp(C*x)))) + B*sqrt(A*A - L*L)*exp(C*x)*sqrt((A*A + 2*A*B*exp(C*x) + B*B* exp(2*C*x) - L*L)/((A + B*exp(C*x))*(A + B*exp(C*x)))) + A*A + A*B*exp(C*x) - L*L))/sqrt(A*A - L*L) + (A*C*x)/sqrt(A*A - L*L))/C ;
```
marcomuzio: break it upppp

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+  double z1 = p->z1;
+
+  struct IceRayTracing::fDnfR_L_params params1a = {A, IceRayTracing::GetB(z1), IceRayTracing::GetC(z1), z1};
+  struct IceRayTracing::fDnfR_L_params params1b = {A, IceRayTracing::GetB(z0), IceRayTracing::GetC(z0), z0};
+  struct IceRayTracing::fDnfR_L_params params1c = {A, IceRayTracing::GetB(IceRayTracing::TransitionBoundary), IceRayTracing::GetC(IceRayTracing::TransitionBoundary), -IceRayTracing::TransitionBoundary};
+  struct IceRayTracing::fDnfR_L_params params1d = {A, IceRayTracing::GetB(-(IceRayTracing::TransitionBoundary+0.000001)), IceRayTracing::GetC(-(IceRayTracing::TransitionBoundary+0.000001)), -(IceRayTracing::TransitionBoundary+0.000001)};
+
+  double distancez0z1=0;
+  if(IceRayTracing::TransitionBoundary!=0){
+    if (fabs(z0)>IceRayTracing::TransitionBoundary && fabs(z1)>IceRayTracing::TransitionBoundary){
```
marcomuzio: I'm kind of confused if this transition boundary is supported. i thought it was commented out from the ice model?

IceRayTracing.cc
```diff
@@ -0,0 +1,3039 @@
...
+
+  vector <double> xD,zD;
+  vector <double> xR,zR;
+  GetFullDirectRayPath_Cnz(z0,x1,z1,lvalueD,A_ice_Cnz,xD,zD);
+  GetFullReflectedRayPath_Cnz(z0,x1,z1,lvalueR,A_ice_Cnz,xR,zR);
+
+}
+
+/* This is the main raytracing function. x0 always has to be zero. z0 is the Tx depth in m and z1 is the depth of the Rx in m. Both depths are negative. x1 is the distance between them. This functions works for a constant refractive index */
+double *IceRayTracing::IceRayTracing_Cnz(double x0, double z0, double x1, double z1, double A_ice_Cnz){
```
marcomuzio: I don't understand this. `x0` needs to be 0? So why is it an argument? Unless you do something like:
```
x1 -= x0;
x0 -= x0;
```
but i dont see that

marcomuzio (changes requested):
Okay I left some comments on the main module that was added. I assume changes in the other files are mostly cleanup.

IceRayTracing.cc
AlanSalcedo: This I will ignore because I don't understand haha

IceRayTracing.cc
AlanSalcedo: It complains about the IceModel function not being static and needed to be called by defining an IceModel object blah blah. 

No way to fix it without messing with IceModel.cc

IceRayTracing.cc
AlanSalcedo: I couldn't get rid of it successfully

PR #210, opened 2025-11-22
Fix secondary cascade readin

abigailbishop:
Increment index that we use to access secondary cascades. Caught because we noticed all secondary interactions were getting saved as the primary cascade.

Currently a work in progress. Currently getting a segfault and there may be a memory leak somewhere.


marcomuzio:
Testing summary
A3 veff test: passed ✅
pa veff test: passed ✅
pulser veff test: passed ✅

PR #213, opened 2026-02-12
Antenna gain translation

marcomuzio:
This PR adds better handling of the translation of antenna gain measurements taken in one medium into another (probably something icy??).

If an antenna gain file does _not_ have an index of refraction, no scaling of the gain argument occurs. 

This PR also removes dependence on a hardcoded `freq_step_max` which sets the frequency resolution. Resolution is now determined dynamically (and so a lot of variables have been updated from arrays to vectors).

This change resulted in a change to the Veff tests, since the normalization of the effective height changed (due to the index in its calculation changed from `n_local` to `n_eff`) and because the frequency resolution of the antenna models changed.

**Testing summary (as of commit 313ad802fa5a13af03c7e9ddce62d923a37be06e)**
A3 veff test: passed ✅
pa veff test: passed ✅
pulser veff test: passed ✅


AlanSalcedo (approved):
Looks good, pending determining what n to use in the normalization of effective height.

Detector.cc
```diff
@@ -2403,8 +2405,45 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
...
-  
+    unique_ptr<istream> streamPtr;
+    ifstream fileTest(filename, ios::binary);
+
+    // Detect gzip by magic bytes
+    unsigned char magic[2] = {0};
+    fileTest.read(reinterpret_cast<char*>(magic), 2);
+    fileTest.close();
+
+    bool isGz = (magic[0] == 0x1f && magic[1] == 0x8b);
```
clark2668: Can you add an explainer for this line?

Detector.cc
```diff
@@ -2403,8 +2405,45 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
...
+        streamPtr = move(f);
+    } else {
+        // Gzip load
+        gzFile gz = gzopen(filename.c_str(), "rb");
+        if (!gz)
+            throw runtime_error("Could not open gz file: " + filename);
+
+        // Decompress entire file into memory
+        string buffer;
+        char tmp[8192];
```
clark2668, line 2433: is this a safe size for tmp?

clark2668 (commented):
Very minor comments

Detector.cc
marcomuzio: Just added a comment explaining this.

Detector.cc
marcomuzio: This is arbitrary, but safe as the loop below reads in the file in chunks of the size of `tmp`. Added a short comment about it to the `while` loop.

clark2668: Okay, thanks.

Detector.cc
clark2668: Thanks

clark2668:
Can we get some plots of the effective height at a couple different zenith angles comparing the measured to the shifted, just to sanity check that all looks as expected?

test/veff_ara3/run_veff_ara3_test.sh
clark2668, line 1: For example, a plot of the heff for this event that gets tipped over into passing might be a good one.

test/veff_pa/run_veff_pa_test.sh
clark2668, line 1: Or this one... Like, this is a pretty significant increase, no?

Constants.h
```diff
@@ -39,6 +39,10 @@ const double Z_AIR=377;            // resistance of air = sqrt(epsilon/mu)
 const double NFIRN=1.3250;                   // index of refraction at the very surface - Peter
 const double NICE=1.79;                      // index of refraction of ice
 
+// see DocDB XXXXX for details
+const double NICE_REF=1.78;           // reference index of refraction for borehole simulation 
+const double NEFF_REF=1.45;           // reference effective index of refraction for borehol simulation
```
clark2668, line 44: I don't understand this. Maybe we need some more docs here.

Detector.cc
```diff
@@ -2417,12 +2455,39 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
...
+        stringstream ss(line);
+
+        // break line up into each of its words (i.e. strings separated by whitespace)
+        string buff;
+        vector<string> words;
+        while(ss >> buff) { // >> skips whitespace and just goes to the next word
+            words.push_back(buff);
+        }
+
+        if(words.size() > 0 && words[0] == "n") { // index included in file
```
clark2668: I don't understand this.

Detector.cc
```diff
@@ -2524,7 +2589,6 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
         getline (NecOut, line);
             
     }// end while NecOut.good
-    NecOut.close();
```
clark2668, line 2527: Don't we need an explicit close?

Detector.cc
```diff
@@ -2634,8 +2698,11 @@ double Detector::GetTransm_OutZero(int ch, double freq) {
 
 }
 
-double Detector::GetGain(double freq, double theta, double phi, int ant_m, int ant_o) { // using Interpolation on multidimentions!
-    
+double Detector::GetGain(double freq, double theta, double phi, int ant_m, int ant_o, double antenna_target_medium_n) { // using Interpolation on multidimensions!
+   
+    // scale frequency according to ratio of media indices of refraction
+    double freq_scaled = (antenna_source_medium_n >= 1)? freq * antenna_target_medium_n / antenna_source_medium_n : freq;
```
clark2668, line 2710: I lowkey hate this >=1 nonsense. Can this be a pure if statement? Or at least have an explainer line?

Detector.cc
```diff
@@ -2881,8 +2951,11 @@ double Detector::GetGain(double freq, double theta, double phi, int ant_m) {
 
 
 
-double Detector::GetAntPhase( double freq, double theta, double phi, int ant_m ) {
+double Detector::GetAntPhase( double freq, double theta, double phi, int ant_m, double antenna_target_medium_n ) {
```
clark2668, line 2966: Should there be a protection (somewhere in the code) to be sure that `antenna_target_medium_n` isn't something stupid?

Detector.cc
```diff
@@ -5263,17 +5347,16 @@ inline void Detector::ReadRayleighFit_TestBed(string filename, Settings *setting
     Rayleigh_TB_databin_ch.resize(ch_no);
     
     // now loop over channels and do interpolation
+    Rayleigh_TB_ch.assign(ch_no, std::vector<double>(freq_step));
```
clark2668: This is just you pre-allocating the memory, right? Can you add a comment?

Detector.cc
```diff
@@ -7071,8 +7154,8 @@ inline void Detector::ReadImpedance(string filename, double (*TempRealImpedance)
             }
 
             impFreq.push_back(stod(words[0]));
-            (*TempRealImpedance)[idx] = stod(words[1]);
-            (*TempImagImpedance)[idx] = stod(words[2]);
+            TempRealImpedance->push_back(stod(words[1]));
```
clark2668, line 7236: Why is TempRealImpedance a pointer, and not just an object? Or maybe it is now? I'm confused.

clark2668 (commented):
Some relatively minor comments.

test/veff_pa/run_veff_pa_test.sh
marcomuzio: No idea. It doesn't seem like a crazy increase to me, especially given the weight is unchanged and the shape of the gain has changed pretty significantly with the increased resolution.

test/veff_ara3/run_veff_ara3_test.sh
marcomuzio: Let me see if I can manage a plot at all first. This isn't easy to get out of AraSim.

Constants.h
marcomuzio: forgot to add the docdb reference in the comment there. thats updated now

Detector.cc
marcomuzio: It's just checking if the line has the index of refraction listed in it, similar to the sanity checks it does on all the lines farther down. The line for the index of refraction should have text and the first column would be `n`. eg: `n : 1.0`

Detector.cc
marcomuzio: It's still a pointer, but I updated it to be a pointer to a vector. It's an alias basically so that this function can be used to load data into several different variables.

marcomuzio:
> Can we get some plots of the effective height at a couple different zenith angles comparing the measured to the shifted, just to sanity check that all looks as expected?

Okay that was painful to get lol. Here's a plot for some random channel. 

Screenshot 2026-02-17 at 1 42 04 PM

Detector.cc
marcomuzio: No it's not a file anymore. It's an `istringstream`: `istream& NecOut = *streamPtr;`

test/veff_pa/run_veff_pa_test.sh
clark2668: What do you mean the weight is unchanged? More events pass, yes

Constants.h
clark2668: Okay

Detector.cc
clark2668: Can you add some extra comments explaining this? That you're looking for "n" because that's how you expect the index to be loaded?

Detector.cc
clark2668: Ahhh, ok.

Detector.cc
clark2668: Ah, okay.

clark2668:
That plot is helpful to see. So it does shift down as expected. And goes up. So this is roughly what we would have anticipated. Okay, thanks.

test/veff_pa/run_veff_pa_test.sh
marcomuzio: The total weight of passing events is essentially unchanged. The new events have extremely small weights.

clark2668: Hmm, I see. Ok.

Detector.cc
marcomuzio: Done.

Detector.cc
marcomuzio: Okay I added a comment.

clark2668: Thank you!

Detector.cc
marcomuzio: Added.

clark2668: Thank you! Did you put it here, or somewhere elsse? GetAntPhase gets called often, a if check maybe slows it down?

Detector.cc
marcomuzio: Done.

marcomuzio:
Okay @clark2668 I'm rerunning the veff tests, but I think I've responded to all your comments.

Detector.cc
clark2668: yay

clark2668:
Okie dokes! Thanks for suffering my review!

Detector.cc
marcomuzio: I put it here. I think there's not a standard place to check this except in all the functions that have that argument directly.

marcomuzio: I didn't see a noticeable slowdown but I didn't really profile it or anything.

clark2668: I mean, okay, good enough for me.

marcomuzio:
Okay all tests have passed. Merging now.

PR #214, opened 2026-02-18
Give receivers and transmitters their own index of refraction

marcomuzio:
This PR gives transmitter and receiver antennas their own indexes of refraction, but still assumes all receivers/transmitters share the same index. This also adds some sanity checks to make sure thats true, but only for receivers since the code currently only supports a single transmitter type at a time.

**Testing summary**
A3 veff test: passed ✅
pa veff test: passed ✅
pulser veff test: passed ✅

### DIFF
--- a/Detector.cc
+++ b/Detector.cc
@@ -2374,6 +2374,7 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
     vector< vector<double> > * gain;
     vector< vector<double> > * phase;
     vector<double>* transAnt_databin;
+    double * source_n;
 
     // make sure dummy variables point to the right variables 
     switch(type) {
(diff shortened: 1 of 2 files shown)


abigailbishop (approved):
Looks okay to me but I'm not the most familiar with what you did

Detector.cc
```diff
@@ -2374,6 +2374,7 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
     vector< vector<double> > * gain;
     vector< vector<double> > * phase;
     vector<double>* transAnt_databin;
+    double * source_n;
```
clark2668, line 2377: does this *need* to be a pointer? I hate pointers.

Detector.h
```diff
@@ -252,6 +252,7 @@ class Detector {
         vector<double> TxFreq;
         vector<vector<double> > Txgain;
         vector<vector<double> > Txphase;
+        double Txantenna_source_medium_n;
```
clark2668, line 255: I'm annoyed this requires a separate variable, but I guess it makes sense. In a different world this could be a property of the antenna. Anyway.

Detector.cc
```diff
@@ -2479,13 +2484,25 @@ inline void Detector::ReadAntennaGain(string filename, Settings *settings1, EAnt
             if(words.size() != 3) {
                 throw runtime_error("Antenna gain file index of refraction not properly formatted! "+filename);
             }
-            antenna_source_medium_n = stof(words[2]);
+
+            double buff_n = stof(words[2]);
+            
+            // always assign if this is the first read-in or its the transmitter
```
AlanSalcedo, line 2490: But you call this function for each antenna type, no? So it's always assigned?

Detector.cc
marcomuzio: yes because its being used as an alias

Detector.h
marcomuzio: right when we refactor this, that's probably how we should do it. cuz all these variables _could_ be different per antenna type, but right now we really only split them between receivers and transmitters.

Detector.cc
clark2668: Okay.

Detector.h
clark2668: Understood. Maybe open an issue?

clark2668:
Okay, lgtm.

AlanSalcedo (approved):
Looks good. Just one question.

Detector.cc
marcomuzio: yes but its a buffer and only assigned to the tracking variables if its different.

AlanSalcedo: Ok

marcomuzio:
Merging.

PR #215, opened 2026-02-18
PA trigger refinement fix.

marcomuzio:
This PR only affects the PA. It enforces that the trigger-level trigger check should pass, not just the high-level trigger checks. 

**Testing summary**
A3 veff test: passed ✅
pa veff test: passed ✅
pulser veff test: passed ✅


abigailbishop (approved):

abigailbishop:
If the PA test passes then I'm happy

PR #221, opened 2026-03-24
Systematics 5sa

AlanSalcedo:
This PR adds the ability to modify the askaryan signal amplitude, attenuation lengths (given a file), and n(z) parameters, so we can evaluate systematics on simulations.

All veff tests passed:

- [ x] veff_ara3
- [ x] veff_birefringence
- [ x] veff_pulser
- [ x] veff_pa


IceModel.cc
```diff
@@ -68,6 +71,22 @@ IceModel::IceModel(int model,int earth_model,int moorebay) : EarthModel(earth_mo
 
   setUpIceModel(model);
 
+  const char* araSimDir = getenv("ARA_SIM_DIR");
```
marcomuzio: I think this is already defined in the header as `ara_sim_dir`

IceModel.cc
```diff
@@ -68,6 +71,22 @@ IceModel::IceModel(int model,int earth_model,int moorebay) : EarthModel(earth_mo
...
+  if (!araSimDir) {
+      std::cerr << "Warning: ARA_SIM_DIR is not set. "
+                << "Ice attenuation systematics table will not be loaded."
+                << std::endl;
+  }
+  else {
+      std::string attenFile =
+          std::string(araSimDir) + "/data/iceattenuation_systematics.txt";
```
marcomuzio: Does this file exist? Or is this just a placeholder for something that needs to be added?

IceModel.cc
```diff
@@ -68,6 +71,22 @@ IceModel::IceModel(int model,int earth_model,int moorebay) : EarthModel(earth_mo
...
+
+      if (!LoadIceAttenPercentTable(attenFile)) {
+          std::cerr << "Warning: failed to load ice attenuation systematics file: "
```
marcomuzio, line 84: Should this throw an error? Or is the idea we always try to load it but maybe we don't always use it?

IceModel.cc
```diff
@@ -2013,11 +2034,91 @@ void IceModel::GetMag (
...
+    iceAttenPctTableLoaded = true;
 
+    std::cout << "Loaded ice attenuation percent table from " << filename
+              << " with " << iceAttenPctDepth.size() << " rows." << std::endl;
+
+    return true;
+}
+
+double IceModel::GetIceAttenSystematicsFactor(double depth, Settings *settings1) const {
+    if (!settings1) {
```
marcomuzio: Is this to make it like an optional argument?

IceModel.cc
```diff
@@ -2013,11 +2034,91 @@ void IceModel::GetMag (
...
+        return 1.0;
+    }
+
+    if (settings1->SYSTEMATICS_IceAttenuation == 0) {
+        return 1.0;
+    }
+
+    if (!iceAttenPctTableLoaded || iceAttenPctDepth.empty()) {
```
marcomuzio, line 2095: Okay maybe now this should be a real crash? I think if you've gotten this far into the function the user is intending to calculate the systematics?

IceModel.cc
```diff
@@ -2013,11 +2034,91 @@ void IceModel::GetMag (
...
+        return 1.0;
+    }
+
+    const double* pct_array = nullptr;
```
marcomuzio: Why not just a vector here?

IceModel.cc
```diff
@@ -2013,11 +2034,91 @@ void IceModel::GetMag (
...
+
+    if (settings1->SYSTEMATICS_IceAttenuation == 1) {
+        pct_array = iceAttenPctUp.data();
```
marcomuzio: then you don't need .data()

IceModel.cc
```diff
@@ -2013,11 +2034,91 @@ void IceModel::GetMag (
...
+        pct_array = iceAttenPctLow.data();
+    }
+    else {
+        return 1.0;
+    }
+
+    double pct = Tools::SimpleLinearInterpolation_extend_Single(
+        (int)iceAttenPctDepth.size(),
+        const_cast<double*>(iceAttenPctDepth.data()),
+        const_cast<double*>(pct_array),
```
marcomuzio: just call .data() here?

IceModel.cc
```diff
@@ -2013,11 +2034,91 @@ void IceModel::GetMag (
...
+        return 1.0;
+    }
+
+    const double* pct_array = nullptr;
+
+    if (settings1->SYSTEMATICS_IceAttenuation == 1) {
+        pct_array = iceAttenPctUp.data();
+    }
+    else if (settings1->SYSTEMATICS_IceAttenuation == 2) {
+        pct_array = iceAttenPctLow.data();
```
marcomuzio: Are these negative numbers?

IceModel.cc
```diff
@@ -249,7 +268,7 @@ if ( file.is_open() ) {
 
 
 // read depth in positive value and return attenuation length (m) at the depth
-double IceModel::GetARAIceAttenuLength(double depth) {
+double IceModel::GetARAIceAttenuLength(double depth, Settings *settings1) {
```
marcomuzio, line 272: Do we also want to add the systematics support to `EffectiveAttenuationLength`?

RaySolver.cc
```diff
@@ -1134,7 +1134,7 @@ void RaySolver::Solve_Ray (Position &source, Position &target, IceModel *antarct
       nc = 0.0132;
     } 
 
-
+    ApplyNofzSystematics(ns, nd, nc, settings1);
```
marcomuzio, line 1137: Doesn't this mean the ice model parameters are incremented in every call to this function? Are they always reset in the function too?

RaySolver.cc
```diff
@@ -1514,6 +1514,11 @@ void RaySolver::Solve_Ray (Position &source, Position &target, IceModel *antarct
 }
 
 
+void RaySolver::ApplyNofzSystematics(double& ns, double& nd, double& nc, Settings* settings1) {
+    if (!settings1) return;
```
marcomuzio: What's the use case here?

Settings.h
```diff
@@ -347,7 +347,16 @@ class Settings
                                              // This setting is only used if DETECTOR > 3 && NOISE == 1 && CUSTOM_ELECTRONICS == 0
                                   
         double CLOCK_ANGLE;  // default: 0; Angle of polarization "on the clock" for use in pulser events (EVENT_TYPE=11)
-
+        
+        int SYSTEMATICS_IceAttenuation; // 0=central, 1=up, 2=low
+        
+        int SYSTEMATICS_Askaryan; // 0=central (default), 1=increase pecentage, 2=decrease pecentage
```
marcomuzio: typos on "percentage"

signal.cc
```diff
@@ -804,6 +804,20 @@ void Signal::Build_Param_RE_Tterm_tables(){
   cout<<"Finished building the Param RE Tterm tables."<<endl;
 }
 
+double Signal::GetAskaryanSystematicsFactor(Settings *settings1) {
+    if (!settings1) return 1.0;
```
marcomuzio: same question

Settings.h
marcomuzio: also do we need this parameter _and_ the one for the size of the change? they seem a little redundant.

marcomuzio (changes requested):
Thanks @AlanSalcedo ! Just a few small things.

IceModel.cc
AlanSalcedo: Added.

Settings.h
AlanSalcedo: It just feels more natural to say 12% and then: add or subtract via this other flag.

AlanSalcedo: Instead of +12% and -12%?

IceModel.cc
AlanSalcedo: Fixed

IceModel.cc
AlanSalcedo: Throwing error

IceModel.cc
AlanSalcedo: Fair enough. Only to the version that already uses has settings in the parameters, which is the only one used in report.cc.

IceModel.cc
AlanSalcedo: This is unnecessary, let me drop it

IceModel.cc
AlanSalcedo: Fair

IceModel.cc
AlanSalcedo: Yes

IceModel.cc
AlanSalcedo: Done

IceModel.cc
AlanSalcedo: Done

IceModel.cc
AlanSalcedo: Done

RaySolver.cc
AlanSalcedo: No, ns, nd, nc are reset in every call, so the increments don't accumulate. That's how it was done already

RaySolver.cc
AlanSalcedo: Dropped

signal.cc
AlanSalcedo: Dropped

Settings.h
AlanSalcedo: Okay, I settled for allowing + or - instead. So we avoid redundancy.

IceModel.cc
marcomuzio: Sorry did the error throw get added?

IceModel.cc
marcomuzio: Sorry to be a pain, I think we should use `throw runtime_error(message);`

RaySolver.cc
marcomuzio: Okay perfect thanks!

signal.cc
```diff
@@ -805,18 +805,17 @@ void Signal::Build_Param_RE_Tterm_tables(){
...
 
-    if (settings1->SYSTEMATICS_Askaryan == 1) {
-        return 1.0 + frac;
-    }
-    else if (settings1->SYSTEMATICS_Askaryan == 2) {
-        return 1.0 - frac;
+    if (factor <= 0.0) {
+        std::cerr << "Error: SYSTEMATICS_AskaryanPercent gives a non-positive scaling factor: "
+                  << settings1->SYSTEMATICS_AskaryanPercent << std::endl;
+        exit(1);
```
marcomuzio: same here

signal.cc
```diff
@@ -805,18 +805,17 @@ void Signal::Build_Param_RE_Tterm_tables(){
...
 
     double frac = settings1->SYSTEMATICS_AskaryanPercent / 100.0;
+    double factor = 1.0 + frac;
 
-    if (settings1->SYSTEMATICS_Askaryan == 1) {
-        return 1.0 + frac;
-    }
-    else if (settings1->SYSTEMATICS_Askaryan == 2) {
-        return 1.0 - frac;
+    if (factor <= 0.0) {
```
marcomuzio, line 812: Why isn't this allowed?

IceModel.cc
marcomuzio: Thanks!

marcomuzio (changes requested):

signal.cc
AlanSalcedo: It would be a scenario where you set the attenuation to < -100%. The factor would be negative, inverting the waveform, but the intention was to attenuate it completely.

IceModel.cc
AlanSalcedo: Done

signal.cc
AlanSalcedo: Done

signal.cc
marcomuzio: Ahh okok. Got it.

marcomuzio (approved):
Thanks Alan!

PR #223, opened 2026-04-11
Fix PA noise sim triggers.

marcomuzio:
This PR fixes an edge case for noise simulations with PA. This only affects noise simulations on PA.

In noise simulations, the PA is forced to trigger and so its trigger bin is `0`. The trigger bin is assigned to `Global_Pass` which is later tested to determine whether there was a trigger via `if (report->stations[i].Global_Pass)`. The bin of `0` is converted to a boolean `false` and so the noise triggers are never considered successful when it comes time to save the event. We fix this by simply returning the center-most available trigger bin (to avoid edge effects) when in noise simulation mode.

**Testing summary**
A3 veff test: passed ✅ 
pa veff test: passed ✅ 
pulser veff test: passed ✅

### DIFF
--- a/Report.cc
+++ b/Report.cc
@@ -5357,7 +5357,7 @@ int Report::get_PA_trigger_bin(
         double avgSnr = 0.;
         if(settings->TRIG_ANALYSIS_MODE == 2) { // Noise only triggers
             avgSnr = pa_force_trigger_snr;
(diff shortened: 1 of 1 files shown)


clark2668:
Ok!

marcomuzio:
Updated to return the center-most trigger bin, since returning `1` was giving nonsensical results (probably due to edge effects).

PR #226, opened 2026-06-05
Adds new ice models for upward and downward systematics.

Machtay:
The new ice models are 41 and 42. They add the systematics from here to the ice model 40 parameters: https://aradocs.wipac.wisc.edu/0035/003525/002/5SA_Ice_Model_Uncertainties.pdf

### DIFF
--- a/RaySolver.cc
+++ b/RaySolver.cc
@@ -1122,6 +1122,16 @@ void RaySolver::Solve_Ray (Position &source, Position &target, IceModel *antarct
       ns = 1.326;
       nd = 1.78;
       nc = 0.0202;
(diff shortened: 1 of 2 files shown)


marcomuzio (changes requested):
Thanks @Machtay this looks good. Can you just add these to the high-level docs in Settings.h? https://github.com/ara-software/AraSim/blob/f6387a5929449c4156810964d00d09964d4c5e31/Settings.h#L287

clark2668:
Agreed about adding to docs, otherwise good.

Machtay:
Added to the docs

marcomuzio (approved):
Thanks!

